PR #33, opened 2026-07-16
feat: route consensus p2p through circuit-relay-v2 (relay-fronted validators)

procdump:
## Problem

The consensus p2p layer only supports **direct QUIC dialing** — validators connect
to each other using the addresses in `committee.yaml`, which exposes every
validator's IP and offers no way to front a validator with a relay, keep it
NAT'd/unreachable directly, or fail over if its ingress point goes away. We want
**circuit-relay-v2 as an option**, so a validator *can* be reached only through a
relay it controls (and isolated / failed over between relays) — while keeping
direct QUIC fully supported. The relay path is opt-in per node via config; nodes
that don't enable it dial directly, exactly as before.

## What's changed

Circuit-relay-v2 is added as an **opt-in** transport path alongside direct QUIC —
enabled per node via keygen/config (`--relay` / `--advertise-dnsaddr` + relay env).
With none of it set, behaviour is unchanged (direct dialing).

**Client (consensus network)**
- circuit-relay-v2 client transport + behaviour: a validator reserves on its relay
  and is reached single-hop through it; outbound dials open circuits to the
  destination's relay.
- PeerManager made relay-aware: relays are exempt from banning/pruning, relayed
  inbound connections are accepted, and committee members are re-dialed on the
  heartbeat so a dropped peer reconnects without waiting for the next epoch.
- Relay reservations are tracked by address and retried (a lost relay is
  re-reserved when it returns) instead of the swarm treating "no listeners" as
  shutdown.

**Relay server**
- new `rayls-relay` circuit-relay-v2 server binary (fixed test identities, raised
  reservation/circuit limits, external-address advertisement so grants carry an
  address, shared QUIC limits with the node, no idle-close of reserving peers).

**DNS / failover**
- validators can advertise a `/dnsaddr` name that resolves (TXT) to *all* of a
  node's relays, with reservations on each, so peers fail over to a backup relay
  when the primary dies. `/dnsaddr` is resolved to concrete `/p2p-circuit`
  addresses at dial time (required for the relay client to classify the connection
  as relayed).

**Testnet tooling & verification**
- `local-testnet.sh --relay` / `--relay-dns`: auto-spawn per-validator relays (and
  dnsmasq for the DNS variant); `add-relay-node.sh` to attach an extra relayed
  node to a running net.
- relay-isolation docker testnet where each validator's *only* neighbor is its own
  relay (relay is default gateway + NAT egress), a topology verifier that proves
  traffic is relayed-only, and a blue-green failover harness with DNS-driven
  cutover.

**Perf**
- announce gossipsub IDONTWANT on publish; apply node QUIC limits on the relay via
  a shared `QuicConfig::apply`.

## Verified
- `--relay` and `--relay-dns` local testnets reach stable consensus with all
  traffic relayed; killing a validator's primary relay keeps consensus running as
  peers fail over to the backup.
- relay-isolation docker net: chain progresses with each validator reachable only
  through its relay (topology verifier confirms no direct validator↔validator
  paths).
- direct-QUIC path (no relay config) unchanged.


crates/middleware/orchestrator/src/epoch_manager/network.rs
```diff
@@ -45,7 +47,35 @@ pub(crate) fn decide_node_mode(
         return match prior_mode {
             NodeMode::CvvActive => (NodeMode::CvvActive, "prior-mode-active"),
             NodeMode::CvvInactive => (NodeMode::CvvInactive, "prior-mode-inactive"),
-            NodeMode::Observer => (NodeMode::Observer, "prior-mode-observer"),
```
procdump: This switch bothers me.

procdump: Do we have to gate this behind the fork as well?

crates/execution/evm/src/evm/hardforks/mod.rs
```diff
@@ -120,7 +120,8 @@ where
             | RaylsHardFork::BatchDigestV2
             | RaylsHardFork::PrecompileGasFix
             | RaylsHardFork::TransactionLoadBalancing
-            | RaylsHardFork::EmptyOutputBlock => continue,
+            | RaylsHardFork::EmptyOutputBlock
+            | RaylsHardFork::DynamicCommitteeSize => continue,
```
procdump: is the committee fix ok to be per block as it should flip at epoch boundary

crates/consensus/network/src/peers/manager.rs
```diff
@@ -418,12 +473,65 @@ impl PeerManager {
...
+    pub(crate) fn register_relays_from_addrs(&mut self, addrs: &[Multiaddr]) {
+        for addr in addrs {
+            if let Some(relay_id) = crate::types::circuit_relay_peer_id(addr) {
+                if self.relay_peers.insert(relay_id) {
+                    debug!(target: "peer-manager", ?relay_id, "registered relay peer (exempt from penalties)");
+                }
+            }
+        }
+    }
+
```
kstoykov: `mark_relay_peer` is making a validator check before inserting in the `relay_peers`. Does it make sense to make such a check in `register_relays_from_addrs`?

crates/consensus/network/src/consensus/gossip.rs
```diff
@@ -108,9 +108,22 @@ where
...
+                // A committee validator reaching here is a real protocol/version fault -- surface
+                // it loudly rather than exempting.
+                if self.swarm.behaviour_mut().peer_manager.mark_relay_peer(peer_id) {
+                    // stop treating it as a DHT peer (mirrors the kad cleanup the ban path did)
+                    self.swarm.behaviour_mut().kademlia.remove_peer(&peer_id);
+                    info!(target: "network", ?peer_id, "peer does not support gossipsub - treating as relay infrastructure (penalty-exempt)");
+                } else {
+                    warn!(target: "network", ?peer_id, "committee member does not support gossipsub - possible protocol/version mismatch");
+                }
             }
```
kstoykov, line 130: Then a peer does not have gossipsub then we mark it as relay peer and we don't apply penalties on it. So anyone without gossipsub is immune to penalties. Can somehow this be exploited?

crates/consensus/network/src/consensus/command.rs
```diff
@@ -314,3 +405,46 @@ where
...
+                    }
+                }
+            }
+            Err(e) => {
+                warn!(target: "network", %name, ?e, "failed to resolve /dnsaddr for relay discovery");
+            }
+        }
+    }
+    circuits
+}
```
kstoykov, line 479: I'm wondering if some can attack us using the DNS. If someone can trick the DNS server then code will just push it in the array as valid. Is there a way to check them against something?

crates/middleware/orchestrator/src/epoch_manager/network.rs
```diff
@@ -312,25 +314,172 @@ where
...
+                .with(Protocol::P2pCircuit)
+                .with_p2p(network_pubkey.clone().into())
+                .map_err(|_| {
+                eyre::eyre!("relay multiaddr from {env_var} ({entry}) has a conflicting P2P id")
+            })?;
+            addrs.push(listen);
+        }
+        Ok(addrs)
+    }
+}
```
kstoykov, line 384: If there is an error this function returns an empty array. Then this array is passed to start_swarm_listeners. It will only log and the swarm will be empty. Therefore we are not starting the swarm in this function but after that we start connecting to peers. Is this how it is supposed to be?

crates/consensus/network/src/consensus/runtime.rs
```diff
@@ -68,6 +111,66 @@ where
...
+            // `retry_relay_reservations`, so an all-relays-down window (a boot race, a
+            // simultaneous relay outage) is waited out instead of shutting the network down.
+            // NOTE: only relay reservations are retried. A node mixing a direct listener with
+            // relay reservations keeps running but never re-establishes the direct listener
+            // (pre-existing behavior restored it via fatal-exit-and-restart); no shipped
+            // topology mixes them today.
+            if self.relay_reservations.is_empty() {
+                error!(target: "network", ?addresses, "no listeners for swarm - network shutting down");
+                return Err(NetworkError::AllListenersClosed);
+            }
```
kstoykov, line 177: Shouldn't this check be only on `active` relays? I mean `.values().any(Some)`? Because now it could not non-empty but contains only Nones.

procdump:
@kstoykov I tried to address your comments - got to the conclusion that I had to lift the ban exemption for relays (which was on the to-do anyway) and make everything to judge by behaviour - at least to me this seems as the ultimate correct goal. You can best review it commit by commit - specifically in regards to each of your comments. Now I'll do some tests to confirm I haven't broken anything which I'll share if so.

kstoykov (approved):

PR #58, opened 2026-07-21
feat: Consensus DB cold storage tiering

bronxyz:
## Summary

The hot consensus MDBX grows without bound (the `Batches` table dominates), so the
shutdown `persist()` flushes scale with uptime instead of staying
flat. But we can't just prune: **the serve floor is genesis** - `backwards_walk` has no
lower bound, catch-up demands every batch digest in every committed sub-dag, and a missing
`Batches` row is a fatal protocol violation. Pruning-without-cold forks any lagging peer.

This adds an **append-only, compressed cold tier** for the consensus DB. Whole archived
epochs of `Batches` and `ConsensusBlocks` relocate into per-epoch
NippyJar files; the hot tables collapse to the
recent window plus a tiny auxiliary index, so the `persist()` barrier stays flat regardless of
uptime. **Nothing is ever dropped** - all history stays online and byte-identically
queryable back to genesis through a read fall-through. The jars commit out of band (own
`sync_all`), never through the consensus `db_run` barrier, so they can't blow the SIGTERM
grace.

## What moves to cold

| table | tier | note |
|---|---|---|
| `Batches` (`digest -> Batch`, the size driver) | **cold**, recent window hot | addressed by the hot `ColdBatchLocations` auxiliary index |
| `ConsensusBlocks` (`u64 -> ConsensusHeader`) | **cold**, recent window hot | dense numbers -> arithmetic addressing, no per-row index |
| `ConsensusBlockNumbersByDigest` | hot | tiny; still grows with chain length |
| `ColdBatchLocations`, `ColdArchiveHighWater` (new) | hot | auxiliary index + the atomicity commit boundary |

## Architecture

- **Read seam (`ColdDatabase<DB>`)** - a `Database` newtype wrapping the hot layer, sitting
  *outermost* above the layered write cache so archival deletes never leave tombstones that
  shadow cold. Reads route `in-mem -> hot -> cold` through the same `Database` trait; every
  age-unbounded read site inherits the fall-through for free. Cold serves are **checked**:
  block number / batch digest are cross-verified against the jar row before returning, so a
  misaligned jar can't serve the wrong row.
- **Cold jars** - one NippyJar per epoch per segment; the `.conf` persists only
  creation-time identity (epoch, start key, kind), and row count / covered range derive from
  the jar itself, so there's no second copy to disagree after a writer heal. `zstd` level 3
  (the compression knee for near-incompressible tx payloads), bounded LRU of open `mmaps`.
- **Crash-atomic ordering** - chunked seal passes (bounded memory, no epoch-long read txn)
  write jars durable -> commit the epoch-folded batch-digest index + high-watermark in one txn ->
  prune hot rows in paced batches. **A row is never absent from both tiers.** `reconcile`
  drives off the `consensus_blocks` jars alone and probes each epoch's last block, healing a
  torn or half-pruned epoch at boot without stranding a tail hot.
- **Background seal actor** - a node-scoped `ColdArchival` aggregate owns boot healing,
  first-start backlog migration, and the seal actor behind one narrow API (inert on backends
  without a cold tier). **Nothing archival runs on the epoch transition** - the actor wakes
  on the executed-anchor epoch flip and seals each due epoch on a dedicated de-prioritized OS
  thread, so archival never competes with consensus. The cutoff keeps only the committed
  epoch hot, floored by the EL execution anchor so the closing epoch stays readable for the
  rewards tally.

## Backwards compatibility & rollout

- **Wire format unchanged.** The only network-facing change routes a peer's batch fetch
  hot-first-then-cold within one read txn; the response is still `Vec<Batch>`. No
  request/response/gossip encoding changes - rolling-upgrade safe.
- **Feature-gated, on by default** (`cold-storage`). Disabling it compiles out the entire
  cold tier and ships the hot-only mdbx stack.

## Operational: `cold-migrate`

A `rayls-network cold-migrate` subcommand (behind the bin's `cold-storage` feature) drains
the boot-time backlog migration (+ optional in-place mdbx compaction) offline, without
starting the node. It shares the boot reconcile + anchor-floored path behind the fail-closed
consensus-DB process lock. Because that lock makes it the sole writer, the layered write queue
never backs up past its high-water mark, so the write pacing never engages; it floors stdout at
`info` so progress lines aren't filtered.

## Operational: `datadir`                                                                                                                                                                                       
Once archival has run, the `cold/` subdirectory of the consensus DB holds the **only** local
copy of every archived `Batches` / `ConsensusBlocks` row. The usual "wipe the consensus DB and
re-sync" recovery must **preserve `cold/`** - deleting it strands history the node is still
expected to serve to lagging peers (a missing `Batches` row is a fatal protocol violation, and
the serve floor is genesis). Back up or move `cold/` alongside the hot DB. 

## Testing

Failing-first proof suite (`crates/infrastructure/storage/src/cold/tests`): hot boundedness
with byte-identical cold serves; a crash-window matrix (torn-between-jar-commits,
pruned-but-unindexed, high-water tail sweep, paced-prune cancel, chunked-seal cancel/retry,
corruption guards); reconcile healing a jar-durable/index-skipped window; the rewards tally
surviving archival at the EL-anchor floor; header-projection == full-decode across header
shapes and single-byte corruptions; a criterion read-path bench.


crates/infrastructure/storage/src/cold/database.rs
```diff
@@ -0,0 +1,313 @@
...
+///
+/// Composed as `ColdDatabase<LayeredDatabase<MdbxDatabase>>` so reads resolve mem -> hot -> cold.
+/// The cold store is shared (`Arc`) because [`Database`] requires `Clone`.
+#[derive(Debug, Clone)]
+pub struct ColdDatabase<DB: Database> {
+    /// The hot database this wraps.
+    inner: DB,
+    /// The shared cold tier consulted on a hot miss.
+    cold: Arc<ColdStore>,
+}
```
kstoykov: I would suggest to wrap the ColdStorage into LayeredDb as it already wraps MemDb and Mdbx.

crates/middleware/orchestrator/src/epoch_manager/engine.rs
```diff
@@ -1,13 +1,47 @@
...
 use tracing::{error, info};
 
+/// Opens the boot EL environment with the exact wiring node boot uses (consistency check and
+/// unwind included).
+///
+/// The single construction point for the boot `RethEnv`: node boot (`create_engine`) and the
+/// offline cold migration both route through it, so the migration's EL anchor is computed against
+/// the same environment a real boot would open. A hand-copied open could silently drift and floor
+/// cold archival differently than boot.
+pub(crate) async fn open_boot_reth_env(
```
kstoykov: What do you think of making this function as part of RethEnv class. Just another `::new_` method?

kstoykov (approved):

PR #63, opened 2026-07-23
feat: apply storage slot fix as hardfork

kstoykov:
## Summary

- Adds a new `GenesisStorageRekey` hardfork that re-keys genesis storage history from plaintext slot keys to hashed keys at activation, fixing a bug where v2 archive reads (e.g. `versions[0]` lookups) returned zero for genesis-seeded storage values
- Removes `#[cfg(feature = "archive-replay")]` guards from `provider_factory` and the `fix_genesis_history*` / `fix_genesis_account_history*` methods so they can be called during live block execution at the hardfork activation block, not only during offline archive replay
- Threads `ProviderFactory` through `RaylsEvmConfig` → `RaylsBlockExecutorFactory` → `RaylsBlockExecutor` so the hardfork migration has access to RocksDB at activation time

Closes #

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

Scheduled at `u64::MAX` for devnet, testnet, and mainnet (not yet activated); block 1 on localnet. Node operators are unaffected until the fork block is set — at that point a coordinated activation is required.

## Test plan

- [x] `cargo test --workspace`
- [x] `cargo test --features archive-replay`

### DIFF
--- a/crates/execution/evm/src/chainspec.rs
+++ b/crates/execution/evm/src/chainspec.rs
@@ -49,6 +49,9 @@ hardfork!(
         /// Produce a fallback empty block for any consensus output that contributed no block
         /// (no batches, all deduped, or all parked), so every output maps to a block.
         EmptyOutputBlock,
+        /// Re-key genesis storage history from plaintext slot keys to hashed keys
+        /// so v2 archive reads reconstruct genesis-seeded storage (e.g. versions[0]).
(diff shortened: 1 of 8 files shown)


kstoykov:
There is a better implementation: https://github.com/raylsnetwork/axyl/pull/64

Issue #48, opened 2026-07-16
Documentation | Create architecture documentation for Rayls Node types (Validator, Observer, Archive)


lucasrparfin:
## Documentation | Create architecture documentation for Rayls Node types (Validator, Observer, Archive)

**Type:** Documentation
**Requested by:** Marcos Viriato

### Context

In the Project Office Status Report thread, Viriato asked that the Rayls node architecture be captured in our documentation, including _why each node type matters for Rayls_. Nuno provided the technical distinctions between node types.

### Objective

Document the three Rayls node types, explaining each one's role, importance, configuration, and relevance to the Rayls network.

### Scope

Covers the three node types running the Rayls mainnet: **Validator**, **Observer**, **Archive**.

### Requirements — the doc must contain

1. An **architecture diagram** showing the node types and how they relate (consensus participation, state, RPC exposure, public access).
2. For each node type: a **definition of its role** and **why it's important**.
3. **Configuration details** per type: consensus participation, state pruning behavior, and RPC exposure (public vs. private).
4. A **"why it matters for Rayls"** section per type

### Key facts to incorporate (from Nuno)

- Validators participate in consensus; RPC-capable but do **not** expose a public RPC → txs cannot be submitted through them.
- Observers do **not** participate in consensus; act as listeners with the same state as validators; expose a public RPC → txs can be submitted.
- Validators and Observers prune state every ~X thousand blocks → pruned blocks can't be replayed and their tx logs are no longer retrievable.
- Archive nodes prune nothing; retain all historical txs and logs; expose a public RPC; typically a must-have for DeFi protocols.
- An Observer can theoretically also act as an Archive node.

### Open questions to confirm before publishing

- [ ] Per-node machine sizing/cost — SRE to provide.

### Definition of Done

- [ ] Diagram included.
- [ ] All three node types documented with role, importance, config, and why-it-matters.
- [ ] Published in docs location (e.g. Notion/Confluence).

luishsr:
PR: https://github.com/raylsnetwork/axyl/pull/70

**Covers:**
- Architecture diagram (Mermaid) — committee membership, consensus role, state streaming, and pruning relationship across the three types.
- Role + "why it's important" for Validator, Observer, and Archive.
- Configuration per type: consensus participation, state pruning, RPC exposure (comparison table + per-section detail, grounded in file:line references).
- "Why it matters for Rayls" per type.
[...]

PR #70, opened 2026-07-24
docs: add node-types page; trim duplicate ASCII diagrams from halt RCA

luishsr:
## Summary
- Adds `doc/node-types.md`: Validator/Observer/Archive reference (role, config, why it matters), cross-linked from `doc/README.md` and `doc/glossary.md`.
- In `doc/incidents/2026-06-21-mainnet-halt.md`, removes the characters-based pipeline/failure/fix diagrams, which duplicated the existing Mermaid diagrams — Mermaid is now the single source. The one unique data point (observed telemetry) is preserved as a plain note.

## Test plan
- [ ] Docs only — no code changes. Preview rendered Markdown/Mermaid on GitHub.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -36,6 +36,9 @@ Start with one of the pages below depending on what you are trying to do.
 - [`index.md`](index.md) — system overview: end-to-end transaction flow,
   architecture diagram, custom database tables, header-field encoding,
   RPC interface.
+- [`node-types.md`](node-types.md) — Validator, Observer, and Archive: what
+  each one is, why it matters, and how consensus participation, state
+  pruning, and RPC exposure differ between them.
 - [`glossary.md`](glossary.md) — terms that mean different things in
   different parts of Axyl (e.g. "network", "whitelist"). Check here before
   using an overloaded term in a design doc or PR description.
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -1,5 +1,8 @@
 # Glossary
 
+> Code pointers below are file-level, not line-pinned — line numbers drift faster than this
+> page gets updated. Last verified against commit `69ee8ab`.
+
 This page exists because a term can mean different things in different parts of Axyl —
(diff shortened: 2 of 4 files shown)


luishsr:
Outstanding from #48's Definition of Done: **per-node machine sizing/cost is still open and needs SRE input** — not addressed by this PR. \`doc/node-types.md\` flags it explicitly under "Open questions" rather than answering it, so it isn't lost, but it should be resolved before this doc is treated as complete guidance for provisioning.

ncervaens:
This documentation helps a lot to understand globally Axyl. As I previously said its very technical with references to repo files so it belong here in .md files.  Its worth that Limechain validates this docs by approving the PR.
But we need to have a more "newbies" documentation in our documentation website. We can create a new task for this.

kstoykov:
I would suggest to remove the link to fixed code lines in the .md files because the code changes quite faster than the .md files. 

Also, do you think it could be good if we add a "Last modified date" or "Valid for commit #...". The idea is that these documents are always behind the code and it could be useful for a reader to know when exactly these docs were updated.

luishsr:
Done in d65bb87 — both suggestions:

- **Dropped line-pinned refs** in `node-types.md` and `glossary.md`: everything is now a file-level pointer (`crates/foo/bar.rs`, not `crates/foo/bar.rs:123`), so a future refactor doesn't leave a misleading line number behind.
- **Added a freshness marker**: both pages now open with "Last verified against commit `<sha>`" right under the title.

Left `doc/incidents/2026-06-21-mainnet-halt.md` as an exception — its file:line citations are a forensic record of the *pre-fix* code at the moment of the incident, not a current pointer, so removing them would lose real information. Added an explicit note there instead, so a reader doesn't mistake them for current line numbers either.

@ncervaens — noted on getting Limechain to review/approve, and agreed a separate "newbies" doc-website task makes sense as a follow-up rather than folding it into this page.

kstoykov (approved):

Issue #79, opened 2026-07-28
Hybrid reward model: activation & rollout (fork schedule, deployment, BlsG1 re-linking)


luishsr:
Follow-up to the hybrid reward **computation** landed in `feat/hybrid-reward-model`. That change ships the reward math only; this issue tracks **activating** it on a network.

## Background

The hybrid `applyIncentives` blend (80% participation / 10% Anchor Commit Share / 10% stake tier) and the off-chain participation tally (`ConsensusHeaderParticipation` + `tally_hybrid`) are implemented and tested, but nothing wires activation: no fork schedule, no change to which contract genesis deploys, no in-place upgrade path. Original issue: raylsnetwork/axyl-private#633.

## Why rollout was deferred (surfaced during review / e2e)

1. **`ConsensusRegistry` deployment/upgrade is non-trivial.** It is a directly-deployed (non-proxy) contract that links the external `BlsG1` library at `owner.create(0)` — a **per-network** address filled in at genesis. A naive bytecode swap carries a wrong, code-less library address and reverts (`stake()` reverts empty). This is e2e-verified, not theoretical. The upgrade must re-link `BlsG1` for the target network, or the network must re-genesis.
2. **Activation ABI must match the deployed contract at every block.** The pre-hybrid contract serves a 1-arg `applyIncentives`; the hybrid contract serves 2-arg. The rollout must keep deployed bytecode and the node's reward-call path consistent across the activation boundary (fresh genesis vs. in-place upgrade), or the first epoch close halts the chain.

## Scope of this follow-up

- [ ] `HybridRewards` hardfork in `chainspec.rs` (`hardfork!` macro, `ForkCondition` schedule per network)
- [ ] `is_hybrid_rewards_active_at_block` + `.hybrid_rewards(block)` builder; update schedule-length assertion
- [ ] `CloseEpochTally` enum threading (`Legacy` / `Hybrid`) in the EVM ctx; `compute_close_epoch_tally` picks `tally()` vs. `tally_hybrid()` by fork; new `ConsensusRegistryV2` `sol!` binding for the 2-arg calldata (leave the 1-arg binding untouched for replay)
- [ ] Bytecode-swap migration (`evm/hardforks/hybrid_rewards.rs`) **with correct per-network `BlsG1` linking**, or a documented re-genesis path
- [ ] Regenerate checked-in `rayls-contracts/artifacts/ConsensusRegistry.json` consumed by genesis
[...]

PR #88, opened 2026-08-03
feat(rewards): HybridRewards hardfork — activate the hybrid reward model

luishsr:
## What

Activates the hybrid reward model (participation + anchor + stake) computed in #80. A new `HybridRewards` hardfork migrates the deployed `ConsensusRegistry` to the hybrid bytecode **in place** and switches the epoch-close reward ABI, gated per network. This is the rollout follow-up tracked in #79.

> **Stacked on #80** (`feat/hybrid-reward-model`) — this PR targets that branch, so the diff is the rollout delta only. Merge after #80; GitHub will retarget it to `main`.

## The deferral blocker, solved

`ConsensusRegistry` links the external `BlsG1` library at a **per-network** address (`owner.create(0)`, filled at genesis) and bakes a per-network `_rls` immutable into its runtime — so no single fixed bytecode can be shipped, the reason activation was deferred from #80.

The migration (`evm/hardforks/hybrid_rewards.rs`) is a **pure byte-splice, no EVM `CREATE`**: it reads the BlsG1 address, `_rls`, and the 5 Solady EIP-712 cached immutables straight out of the **live contract's runtime code** and splices them into the checked-in hybrid bytecode fixture, then installs it preserving storage/nonce/balance. Why it's sound:
- The 5 EIP-712 immutables keep **stable AST ids** across the pre-hybrid→hybrid compilations (only `_rls` shifts), so they map 1:1.
- `nameHash`/`versionHash` are unchanged constants; `cachedThis`/`cachedChainId`/`cachedDomainSeparator` self-heal at call time (Solady recomputes when `address(this)`/`chainid` disagree with the cached values).
- `offsets_match_live_artifact` re-derives the old offsets from the checked-in artifact and fails if the hardcoded constants ever drift.

## Fork flag

`HybridRewards` (version `0x0d`) across the `hardfork!` macro, version byte, and all four schedules — **Local `Block(1)`** (genesis-old → migrate-at-first-block, exercised by tests), **Devnet/Testnet/Mainnet `Never`** until SRE schedules a concrete block — plus the `is_hybrid_rewards_active_at_block` accessor, a `.hybrid_rewards(N)` builder, and the length-assertion test.

## ABI selection

- A separate **`ConsensusRegistryV2`** `sol!` binding for the 2-arg `applyIncentives(RewardInfo[], totalRounds)`, leaving the 1-arg encoding byte-identical for archive replay.
- A **`CloseEpochTally`** `Legacy`/`Hybrid` enum — the single source of truth for which ABI encodes, so the fork check and calldata shape can't disagree.
- `compute_close_epoch_tally` picks `tally()`/`tally_hybrid()` by the **close block's number** → a clean whole-epoch cutover (epochs are timestamp-delineated but each closes in exactly one block; no epoch is split).
- `Withdrawal.amount` keeps meaning leader/anchor rounds post-fork (replay compatibility).

## Tests

- **Migration units (5):** splice round-trip, unspliced-bytes-unchanged, wrong-length reject, fixture length/zeroed-links, and `offsets_match_live_artifact`.
- **`test_hybrid_rewards_fork_boundary`:** a synthetic `.hybrid_rewards(3)` schedule drives a pre-fork close (legacy 1-arg vs. the old contract) and a post-fork close (migration + hybrid 2-arg) in one run — a wrong ABI or bad migration would revert `applyIncentives` and fail the block. Registry code length flips **21929 → 23042** as direct migration proof.
- **`test_close_epochs`** already executes the migrated contract on Local (hybrid at block 1).
- Chainspec suite (14 tests) updated for the new fork.

## e2e

`test_epoch_boundary` is wired for the hybrid path (`--network local`), and `get_rayls_network_binary` now honors `CARGO_TARGET_DIR` so escargot can build from a space-free path (jemalloc-sys rejects spaces in the build path — a portability fix for any space-containing checkout). **The multi-node run is currently blocked by a pre-existing, unrelated `invalid chain ID` failure in that `#[ignore]` test** — confirmed to reproduce with none of these changes active. The hybrid path itself is validated end-to-end by the in-process executor tests above.

## Not included (follow-ups)
- The concrete mainnet/testnet activation block (SRE decision; `Never` until then).
- Fixing the pre-existing `test_epoch_boundary` chain-id bug to enable the multi-node hybrid e2e.
- Archive replay of hybrid epochs (`SnapshotRewardsBackend::tally_hybrid` hard-errors by design, from #80).

Addresses #79.


crates/execution/evm/src/evm/hardforks/hybrid_rewards.rs
```diff
@@ -0,0 +1,287 @@
...
+        })?;
+
+    let live_code: Bytecode = match info.code {
+        Some(code) => code,
+        None => db.code_by_hash(info.code_hash).map_err(|e| {
+            BlockExecutionError::msg(format!("HybridRewards: load ConsensusRegistry code: {e}"))
+        })?,
+    };
+
+    let new_code = splice_hybrid_registry_code(live_code.original_byte_slice())?;
```
bronxyz, line 166: Q: Wouldn't setting the code from the already compiled binary here not work, without having to do the splicing "by hand"?

luishsr: No — the artifact's `deployedBytecode` isn't usable as-is, and genesis doesn't use it either.

Two classes of value are baked into the **runtime bytes** (not storage), and both are per-network:

1. **BlsG1 library address** — `ConsensusRegistry` calls an external library, so solc leaves 8 link placeholders in the runtime code. The address is `owner.create(0)`, so it differs per network.
2. **Immutables** — `_rls` (7 sites) plus Solady's 5 cached EIP-712 values. Solidity `immutable`s live in code, not storage; the constructor patches them into the runtime bytes at deploy time. The checked-in `deployedBytecode` has all of them zeroed.

Install those bytes verbatim and you get a registry whose `_rls` is `address(0)` and whose BlsG1 calls target `address(0)` — `stake()` / `safeTransferFrom` revert and staking is dead.

That's also why genesis doesn't use `deployedBytecode`: it links BlsG1 into the *init*code, runs the real constructor via `transact_pre_genesis_create` on a throwaway chain, and lifts the **resulting** runtime code into the genesis alloc (`reth_env/genesis.rs:283-300`, `:495`).

So for the migration there are two options:

- **Re-run a `CREATE`.** Migrations today are pure state deltas with no EVM execution, so this means threading the EVM factory into the migration dispatcher — and the constructor also re-initialises validator/epoch storage at a fresh address, so we'd have to copy storage across afterwards. Much more machinery, and much more that can silently go wrong on a live chain.
- **Carry the per-network values over from the live code** — the splice. Deterministic, no EVM, and address/storage/nonce/balance are untouched by construction.

It's the same operation solc's linker does for libraries (see `link_solidity_library` in `genesis.rs`), extended to immutables. And it's guarded rather than trusted: length check, cross-site agreement for BlsG1 and `_rls`, `offsets_match_live_artifact` re-derives every old offset from the checked-in artifact (now including the EIP-712 ones, plus a pin on the total immutable count so a newly-added immutable can't be silently left zeroed), and `fixture_has_expected_length_and_zeroed_splice_sites` asserts every site the splice writes is zero in the fixture — so a drifted offset can't partially overwrite logic bytes.

One more constraint worth noting: unlike the five UUPS precompile proxies, `ConsensusRegistry` is deployed directly at a fixed address with no proxy in front of it, so there's no `upgradeTo` path — replacing the code in place is the only upgrade mechanism available.

luishsr:
Thanks — addressed. Summary:

**1. `OLD_EIP712_OFFSETS` not validated — fixed.** `offsets_match_live_artifact` now derives them from the artifact. One deliberate change from the suggestion: rather than sorting both sides, it sorts the single-site immutables by **AST id** and compares the `start` values **in that order**. The splice pairs old→new positionally by ascending AST id, so sorting both sides by offset would let an ordering swap through — verified by perturbation (swapping the first two entries fails the test as written; with sort-both it would pass). Also added a pin on the total immutable count (`_rls` + 5 EIP-712), so a newly-added immutable in a recompiled contract can't be silently left zeroed by the splice.

**2. `NEW_EIP712_OFFSETS` zeroed check — fixed.** Added, and also for `NEW_RLS_OFFSETS`, which had the same gap. Renamed the test `..._zeroed_link_sites` → `..._zeroed_splice_sites` since it now covers every site the splice writes. Also verified by perturbation.

**3. Indexed loop → `zip` — fixed.** Plus `const _: () = assert!(OLD_EIP712_OFFSETS.len() == NEW_EIP712_OFFSETS.len());` next to the constants.

**7. Version-byte test — fixed.** `local_network_version_byte_advances_at_hybrid_rewards` asserts `0x0c` at `LOCAL_HYBRID_REWARDS_BLOCK - 1` and `0x0d` at the activation block and beyond.

**4. Gating `apply_reward_distribution` on Tokenomics — not doing this; the suggested fix would break mainnet.** `Tokenomics` is `ForkCondition::Never` on mainnet (`chainspec.rs:270`), so gating the call on `is_tokenomics_active_at_block` would disable reward distribution on mainnet entirely and change the state root of every epoch-close block → chain split. The premise is also off: `RewardDistributor` **is** deployed on Local — genesis allocates its UUPS proxy unconditionally for every network (`reth_env/genesis.rs:546-551`). And the call site is untouched by this PR; the only change in that region is the `tally` type. If the log noise is real it's worth its own issue, but the fix would need its own fork gate, not a bare condition.

**5, 6** — agreed, no change needed.

Verification: all 88 `rayls-execution-evm` lib tests pass (including `test_hybrid_rewards_fork_boundary` and `test_close_epochs`), `cargo clippy` clean, `cargo +nightly fmt --all --check` clean. Formatting also picked up four pre-existing spots in this PR's own code — `pr.yaml` hasn't run on this branch because it targets `feat/hybrid-reward-model` rather than `main`.

bronxyz (dismissed):

bronxyz (approved):

PR #92, opened 2026-08-05
fix: unify the ban predicate across admission and disconnect

bronxyz:
## Summary

Align a peer's connection state with its reputation verdict on the admission and disconnect paths.

- disconnecting at the score threshold no longer bans the peer: a Disconnected verdict maps to DisconnectReason::Penalized, which neither sets ConnectionStatus::Banned nor charges the peer's ips into the ban table
- register_peer_connection gates on PeerManager::peer_banned, the same predicate the admission gates use, so a committee member admitted at the gate is not then refused at registration
- a connection register_peer_connection refuses is closed, so the swarm does not keep a connection the manager holds no state for

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

### DIFF
--- a/crates/consensus/network/src/peers/all_peers.rs
+++ b/crates/consensus/network/src/peers/all_peers.rs
@@ -4,8 +4,12 @@
 //! manager to take. Some actions are propagated up to the swarm level and affect other behaviors.
 
 use super::{
-    banned::BannedPeers, peer::Peer, score::ReputationUpdate, status::ConnectionStatus,
-    types::ConnectionDirection, PeerExchangeMap, Penalty,
+    banned::BannedPeers,
+    peer::Peer,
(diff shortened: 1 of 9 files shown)


bronxyz:
- Stale `INVARIANT`/`CURRENT` comments: addressed in #98. The suite header now frames the pairs as `PRE-FIX` regression guards, since the pinned bugs are fixed and the tests pass.
- `assert_ban_accounting_conserved` defined but never called: stacked-PR artifact. It is called by the tests added in #93 and later; reviewing this PR in isolation hides the callers.
- `score_or_ip_banned` "single caller" doc wording: kept as is. The sentence is an intentional caution against gating a connection path on the raw predicate.

kstoykov (approved):

PR #93, opened 2026-08-05
fix: conserve ban-table accounting on ban and release

bronxyz:
## Summary

Make the ban-table accounting a conserved quantity that cannot drift from the set of banned peers.

- capacity eviction retires the oldest bans instead of the newest, so a flood of new identities cannot flush the bans just earned
- a ban is released against exactly the ip addresses it charged, keyed by peer id; the total is derived from that map rather than a hand-maintained counter, so it cannot drift

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

### DIFF
--- a/crates/consensus/network/src/peers/all_peers.rs
+++ b/crates/consensus/network/src/peers/all_peers.rs
@@ -91,7 +91,7 @@ impl AllPeers {
     ) {
         let peer_id: PeerId = network_key.clone().into();
         let trusted_peer = Peer::new_trusted(bls_public_key, network_key, addr);
-        let _ = self.banned_peers.remove_banned_peer(trusted_peer.known_ip_addresses());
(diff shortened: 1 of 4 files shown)


bronxyz:
- `remove_banned_peer` reports count 1->0 IPs as unban events: consciously skipped. Pre-existing behavior and a harmless no-op, since unbanning an IP that was never blocklisted does nothing.
- `Unban` action uses current IPs rather than charged IPs: consciously skipped. Pre-existing and separate from this PR's bug; the charged-IP set is now recorded, so a later change can switch to it if wanted.
- Decay overshoot / sleep-based tests: informational; the tests pass under the CI runner.

kstoykov (dismissed):

kstoykov (approved):

PR #94, opened 2026-08-05
fix: decay on wall time and act on the heartbeat verdict

bronxyz:
## Summary

Correct reputation decay and ensure every heartbeat verdict is acted on.

- decay accounts the fractional part of the elapsed interval, so a caller running more than once a second no longer stops decay entirely and strands a ban forever
- a committee member whose record resolves after the epoch boundary is recognized immediately, restoring its penalty absolution and ban exemption instead of waiting a whole epoch
- the heartbeat acts on the disconnect verdict it reaches instead of dropping it and restating it as an error every beat

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

### DIFF
--- a/crates/consensus/network/src/peers/all_peers.rs
+++ b/crates/consensus/network/src/peers/all_peers.rs
@@ -103,21 +103,30 @@ impl AllPeers {
         addrs: Vec<Multiaddr>,
     ) {
         let peer_id: PeerId = network_key.clone().into();
+
+        // Check if this peer is a committee member - only committee members get trusted status
+        let is_committee_member = self.current_committee_keys.contains_key(&bls_public_key);
+        if is_committee_member {
+            // this call is the only place the key-to-peer mapping becomes known for a member the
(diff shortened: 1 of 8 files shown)


bronxyz:
- Banned committee member trusted by discovery keeps its ban charge: confirmed and fixed in #98. `add_known_peer` releases the ban once the resolved member is a validator, with a failing-first regression test. The release lives in the discovery path, not `upsert_peer`, because `new_epoch` also reaches `upsert_peer` and performs its own boundary unban.
- Global score-config `RwLock` overwrite races under parallel tests: consciously skipped. CI mandates `cargo test -- --test-threads 1` (pr.yaml, with an explicit not-nextest note), and the constraint is documented at the config site.
- `upsert_peer` double-insert into `current_committee`: consciously skipped; the insert is idempotent.

kstoykov (dismissed):

kstoykov (approved):

PR #95, opened 2026-08-05
fix: charge ip bans from observed state and unify thresholds

bronxyz:
## Summary

Ground ban-ip charges and thresholds in observed state, and propagate a cleared ban to the stores that hold it.

- the ip ban table is charged only from addresses this node observed the peer connect from, not from addresses the peer advertised in its own record
- a ban cleared by a reconnect or a dial returns PeerAction::Unban, so the gossipsub blacklist and kad routing entry the ban removed are repaired
- reputation reads one configured pair of thresholds, so the decay freeze and the ban verdict describe the same peers

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

### DIFF
--- a/crates/consensus/network/src/peers/all_peers.rs
+++ b/crates/consensus/network/src/peers/all_peers.rs
@@ -390,6 +390,11 @@ impl AllPeers {
         multiaddr: Multiaddr,
         direction: ConnectionDirection,
     ) -> PeerAction {
+        // a ban cleared here has to be announced: the gossipsub blacklist and the kad routing
+        // entry are repaired only by `PeerAction::Unban`, so releasing the counters silently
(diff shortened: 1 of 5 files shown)


bronxyz:
- `update_net` double-negation doc: fixed in #98.
- `reputation()` panic path: not an issue; it is the same must-be-initialized precondition `Score::default` and `apply_penalty` already carry.
- `exchange_info` no longer propagating peer-advertised addresses: consciously kept. It is the point of the witnessed-vs-advertised split: a peer must not be able to charge or advertise an address this node never observed. Peers behind NAT stay discoverable through their signed KAD records.

kstoykov (dismissed):

kstoykov (approved):

PR #96, opened 2026-08-05
fix: correct committee handover, ban logging, and kad penalty

bronxyz:
## Summary

Correct committee handover, surface the ban lifecycle, and attribute kad record faults to their author.

- a committee member departing at an epoch boundary keeps gossip-authorization importance for one epoch, but loses its penalty absolution and ban exemption at the boundary
- the ban lifecycle is logged at warn and info from the single point every reputation action converges, so an operator sees the ban and the score that caused it
- a kad record fault is charged to the record's author: a content-integrity fault always charges the deliverer, an author-attributed fault charges only the named author, so an honest relayer earns no penalty

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

### DIFF
--- a/crates/consensus/network/src/consensus/gossip.rs
+++ b/crates/consensus/network/src/consensus/gossip.rs
@@ -73,8 +73,10 @@ where
                         author = ?gossip_source,
                         ?topic,
                         ?propagation_source,
-                        "applying fatal penalty to message author (or propagation source as fallback)"
+                        "applying fatal penalty to message author"
                     );
+                    // Penalize the message author only, never the forwarding propagation source; a
+                    // relayer must not be banned for an invalid message it merely gossiped on.
                     if let Some(peer_id) = gossip_source {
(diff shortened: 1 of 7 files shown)


bronxyz:
- `ValueTooLarge` charging the deliverer: confirmed and fixed in #98. The store error path now sees the authenticated `publisher` and charges Fatal only when the deliverer authored the record, matching the deliverer-vs-author invariant.
- Dead `MaxProvidedKeys` arm: removed in #98. Provider records are dropped before `add_provider`, so the arm is `unreachable!()` and the dead `RecordInvalidReason` variants are deleted.
- `MaxRecords` store-full penalty converging on disconnects under load: consciously deferred. The trigger is the local store filling, a separate hardening from record-fault attribution.

kstoykov (dismissed):

kstoykov (approved):

PR #99, opened 2026-08-05
feat(observability): enable metrics via config + operator monitoring setup

luishsr:
## What

The **"turn on & unify" P1** of the Network Observability Initiative (`axyl-private#429`, research task `#381`). The research audit found the node is **already heavily instrumented** — it inherited Narwhal/Mysten's full Prometheus suite (consensus/primary/worker/network/executor metrics) *and* has reth's execution metrics — but everything is **off by default**, **split across two registries**, with **no config-file route** to enable it. This PR closes the turn-on & unify gaps.

## Config — enable metrics from `parameters.yaml`

Both endpoints are now enable-able from config, not just CLI flags:

```yaml
metrics_address: "0.0.0.0:9184"        # consensus / Narwhal suite (--metrics)
reth_metrics_address: "0.0.0.0:9001"   # reth execution layer (--reth-metrics)
```

- Both `Option<SocketAddr>` with `#[serde(default)]` — **every existing `parameters.yaml` still parses and stays off** (covered by tests).
- The `--metrics` / `--reth-metrics` **CLI flags override** the config values when passed (mirrors the existing `--network` override); startup logs announce each active endpoint.
- Removes the **dead `PrometheusMetricsParameters`** struct — defined with a `Default` impl but wired nowhere, and typed `Multiaddr` where the live path is `SocketAddr`.
- Fixes a stale `--enable-healthcheck` doc reference (the real flag is `--healthcheck`).

## Monitoring — `etc/monitoring/`

A ready-to-run operator stack:
- **`prometheus.yml`** — a **two-target scrape** (consensus + execution + `node_exporter`), labeled by `layer`, which resolves the split-registry issue (the two endpoints live in separate registries, so they're scraped as separate jobs).
- **`docker-compose.yml`** — one-command Prometheus + Grafana + node_exporter.
- Grafana datasource auto-provisioning + a **README** walkthrough pointing operators at reth's official [Grafana dashboard #20638](https://grafana.com/grafana/dashboards/20638-reth/).

## Tests

4 `config::node::tests` cover backward-compat (no field → off), parsing when present, and the defaults for both endpoints. Compiles clean.

## Scope / next

This is turn-on & unify only. The **validator-identity metrics layer** (per-validator liveness/participation, committee/epoch/reputation), a **real readiness probe**, and a **Rayls consensus/validator Grafana dashboard** are the next deliverable, tracked in `#428`. The full research (current-state audit + reuse-vs-build) is attached to `#381`.

### DIFF
--- a/crates/execution/evm/src/reth_env/config.rs
+++ b/crates/execution/evm/src/reth_env/config.rs
@@ -29,6 +29,8 @@ use tracing::warn;
 #[derive(Debug, Clone, Default, Parser)]
 pub struct RethMetricArgs {
     /// Enable Prometheus metrics for reth execution-layer components.
+    ///
+    /// Overrides `reth_metrics_address` in parameters.yaml when passed.
     #[arg(long = "reth-metrics", value_name = "SOCKET", value_parser = parse_socket_address, help_heading = "Reth Metrics"
     )]
     pub prometheus: Option<SocketAddr>,
--- a/crates/infrastructure/config/src/node.rs
+++ b/crates/infrastructure/config/src/node.rs
@@ -2,13 +2,12 @@
 
 use crate::{ConfigFmt, ConfigTrait, NodeInfo, RaylsDirs};
 use rayls_infrastructure_types::{
-    get_available_tcp_port, get_available_udp_port, test_genesis, Address, BlsPublicKey,
-    BlsSignature, Genesis, Multiaddr, NetworkPublicKey, RaylsNetwork,
-    ETHEREUM_BLOCK_GAS_LIMIT_56BITS, MIN_RAYLS_PROTOCOL_BASE_FEE,
+    get_available_udp_port, test_genesis, Address, BlsPublicKey, BlsSignature, Genesis,
+    NetworkPublicKey, RaylsNetwork, ETHEREUM_BLOCK_GAS_LIMIT_56BITS, MIN_RAYLS_PROTOCOL_BASE_FEE,
 };
 use reth_chainspec::ChainSpec;
 use serde::{Deserialize, Serialize};
-use std::time::Duration;
+use std::{net::SocketAddr, time::Duration};
 use tracing::info;
 
 /// The filename to use when reading/writing the validator's BlsKey.
@@ -216,6 +215,15 @@ pub struct Parameters {
     /// Controls the maximum gas per block and batch.
     #[serde(default = "Parameters::default_gas_limit")]
     pub gas_limit: u64,
+    /// Address for the consensus Prometheus metrics endpoint (the Narwhal/consensus metrics
+    /// suite). `None` (the default) leaves it off. The `--metrics` CLI flag overrides this when
+    /// passed, so operators can enable metrics from `parameters.yaml` without a flag.
+    #[serde(default)]
+    pub metrics_address: Option<SocketAddr>,
+    /// Address for the reth execution-layer Prometheus metrics endpoint. `None` (the default)
+    /// leaves it off. The `--reth-metrics` CLI flag overrides this when passed.
+    #[serde(default)]
+    pub reth_metrics_address: Option<SocketAddr>,
 }
 
 impl Parameters {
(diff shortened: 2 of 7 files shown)


bronxyz:
Thanks for the research done!
The clean-up part for the `PrometheusMetricsParameters` and the docs tidying are good.
We already expose both consensus and reth metrics in all of our scripts via `--metrics 0.0.0.0:9100` and `--reth-metrics 0.0.0.0:9200` and currently the entire stack is configured around that.
Let's discuss how to proceed with this.

luishsr:
Thanks @bronxyz — and good context on the scripts.

One clarification on how this interacts with the flag-based setup, since it changes the picture: **the CLI flags take precedence over config.** In `network-cli/src/node.rs` the resolution is `metrics.or(parameters.metrics_address)` (same for reth), so whenever `--metrics` / `--reth-metrics` are passed they win, and the config is only consulted when the flag is absent. Both fields default to `None` (off), and existing `parameters.yaml` files parse unchanged. So your scripts and the whole `9100`/`9200` stack are unaffected regardless of what we decide here — there's no competing source of truth.

Given that, the only open question is whether the config path is worth keeping as an additive option:

- **Keep it (additive default):** lets a deployment enable metrics declaratively in the same `parameters.yaml` it already uses for `gc_depth`/`gas_limit`/etc., without every launch path having to remember the flags. Directly addresses the initiative's #1 finding ("metrics are off unless a flag is passed"). Costs a little config surface area.
- **Flags-only (descope the two fields):** keep the dead-code cleanup + docs + the `etc/monitoring` operator stack you're happy with, and drop `metrics_address` / `reth_metrics_address` since every launch path already sets the flags. Less surface; we can revisit if a non-script launch path ever needs it.

I'm happy either way — which do you prefer? If flags-only, I'll strip the two config fields and keep the rest.

bronxyz:
I'd say we can keep the params as fallback, but let's align the ports with the script defaults, so we can avoid any potential ports collision.

procdump:
@luishsr , can you confirm that these changes won't break the local-network test environment? maybe deployment scripts as well?

luishsr:
Checked — no break to the local-network test env or deployment scripts:

- **The new config fields are purely additive.** `metrics_address`/`reth_metrics_address` on `Parameters` are `#[serde(default)]`, defaulting to `None`. I checked the actual local-network fixtures (\`etc/test-network/local-validators/{,observer,validator-1..4}/parameters.yaml\`) — none of them set these keys today, so they parse identically before/after this PR and metrics stay off, same as now. There's also an explicit test (\`parameters_without_metrics_address_default_to_off\`) covering exactly this case.
- **CLI flags are unchanged in name and type.** \`--metrics\` and \`--reth-metrics\` are the same flags as before (only the doc text improved); this PR only adds a config-file fallback (\`metrics.or(parameters.metrics_address)\`), and the CLI value always wins when passed — so any deployment script that already passes \`--metrics <addr>\`/\`--reth-metrics <addr>\` explicitly keeps working exactly as before.
- **`PrometheusMetricsParameters` (removed) was dead code.** Grepped all of \`origin/main\` — it's referenced nowhere outside its own definition (not a field on `Parameters`, not constructed or read anywhere), so removing it doesn't change any runtime behavior for existing configs.
- The `health.rs` doc-comment line in this diff is the same pre-existing-stale-doc fix I dug into on #102 — the actual `--healthcheck <PORT>` flag itself is unchanged there too (see my reply on that thread for the full trace).

So: nothing here should affect the local test environment or existing deployment scripts. Let me know if there's a specific script/config outside this repo you'd like me to check against directly.

procdump:
can you fix the fmt problem the CI reports and also rebase?

kstoykov:
I would just suggest to pin the versions in etc/monitoring/<docker-compose> files to a specific version.

luishsr:
Rebased and pushed (4bbd193, on top of latest main) with all three items addressed:

- **Fixed the fmt failure** CI reported (an import-list wrapping issue in `crates/infrastructure/config/src/node.rs`), verified against the exact pinned CI nightly (`nightly-2026-06-24`), not just a local generic nightly.
- **@bronxyz** — aligned the `etc/monitoring/` example ports with what your scripts already use (`--metrics 0.0.0.0:9100`, `--reth-metrics 0.0.0.0:9200`), replacing the 9184/9001 placeholders. While doing that I caught a real collision the old placeholders had been hiding: `node-exporter`'s own default port is *also* 9100, so publishing it on the host at 9100 would have collided with a node's `--metrics 0.0.0.0:9100` on the same box. Moved node-exporter's host-side mapping to 9101 — Prometheus's internal scrape of it (`node-exporter:9100`, a docker-network address) is unaffected.
- **@kstoykov** — pinned the three images: `prom/prometheus:v3.8.1`, `grafana/grafana:12.1`, `prom/node-exporter:v1.10.2` (checked current stable tags rather than guessing).

`cargo check` on the touched crates passes clean post-rebase.

procdump (dismissed):

procdump:
@luishsr rebase please

kstoykov (dismissed):

kstoykov (approved):

PR #102, opened 2026-08-06
feat(observability): validator health metrics + readiness probe

luishsr:
## What

**P2 of the Network Observability Initiative** (`axyl-private#428`). The research audit found the node already exports Narwhal/Mysten's full Prometheus suite — but the highest-value **per-validator** and **readiness** signals were tracked in-memory and never exposed. This PR surfaces them.

## Readiness probe

The healthcheck now serves **`/readyz`** — `200` only when the node is **voting (`CvvActive`)** or an operational observer, and **`503` while a validator is still catching up (`CvvInactive`)**. Liveness (any other path) stays `200`. It reads the live `NodeMode` via a `watch::Receiver` from the consensus bus, threaded in at the spawn site. This lets a load balancer / on-call distinguish **"process up"** from **"actually voting in the current epoch"** — which the unconditional-`200` healthcheck couldn't.

## New metrics (consensus registry, served on `--metrics` / `metrics_address`)

- **`validator_participation{authority}`** — committed certificates per validator, counted at the bullshark commit site (where the sub-dag certificates are already iterated). A **direct per-validator liveness signal**, and **consistent across nodes** since every node commits the same sub-dags. *This is the audit's biggest gap — "did validator X vote/certify?"*
- **`vote_request_rejections{authority, reason}`** — this node's header vote requests rejected by a peer (from `VoteFailureTracker`), by rejecting validator and reason (`too_old` / `epoch_mismatch`). Node-local: rising totals mean this node is falling behind. Complements the network-consistent `validator_participation`.
- **`committee_size`** — committee size for the current epoch, set at the per-epoch metrics hook.

## Dashboard

`etc/monitoring/.../rayls-consensus.json` — a starter Grafana dashboard (9 panels: epoch, committee size, bad nodes, round progress, commit-latency p95, leader election, the two new per-validator series, peers) with auto-provisioning, portable via a datasource variable.

## Tests

3 tests on the readiness probe (path routing, mode→status mapping, and a live server test flipping `CvvInactive`→`CvvActive` and seeing `503`→`200`). The metrics follow the established `register_*_with_registry!` idiom; compiles clean across `consensus-primary`, `primary-metrics`, and `orchestrator`.

## Composition / scope

- The `etc/monitoring/` dashboard **composes with the monitoring stack in #99** (this branch is off `main`, so #99's `prometheus.yml` / docker-compose aren't on it yet; they merge cleanly as separate files).
- Deferred follow-ups: on-chain `ValidatorStatus` export (Active / PendingActivation / Exited — needs an EVM `ConsensusRegistry` read joined via `Committee`) and round-lag as its own metric (currently derived in the dashboard). Tracked under `#428`.

### DIFF
--- a/crates/middleware/orchestrator/src/types/health.rs
+++ b/crates/middleware/orchestrator/src/types/health.rs
@@ -3,10 +3,10 @@
 //! A minimal HTTP/1.1 server exposing two probes:
 //! - **liveness** (any path, e.g. `GET /`): always `200 OK` while the process is accepting
 //!   connections — the pre-existing behaviour.
-//! - **readiness** (`GET /readyz`): `200 OK` only when the node is actually participating —
-//!   voting (`CvvActive`) or an operational observer — and `503 Service Unavailable` while a
-//!   validator is still catching up (`CvvInactive`). This lets a load balancer / on-call
-//!   distinguish "process up" from "actually voting in the current epoch".
+//! - **readiness** (`GET /readyz`, alias `GET /ready`): `200 OK` only when the node is actually
+//!   participating — voting (`CvvActive`) or an operational observer — and `503 Service
+//!   Unavailable` while a validator is still catching up (`CvvInactive`). This lets a load
+//!   balancer / on-call distinguish "process up" from "actually voting in the current epoch".
 //!
 //! Designed for integration with GCP load balancers and similar health monitoring systems.
 use std::{io::ErrorKind, net::SocketAddr, time::Duration};
@@ -81,32 +81,46 @@ impl HealthcheckServer {
         info!(target: "epoch-manager", ?listen_on, "healthcheck listening");
 
         task_spawner.spawn_task("healthcheck", async move {
-            let mut backoff = Duration::from_millis(100);
+            let initial_backoff = Duration::from_millis(100);
+            let mut backoff = initial_backoff;
             let max_backoff = Duration::from_secs(5);
 
             loop {
                 match listener.accept().await {
                     Ok((mut socket, _)) => {
-                        // Read the request line to route by path. Bounded read with a short
-                        // timeout so a slow/partial client can't wedge the accept loop.
-                        let mut buf = [0u8; 256];
-                        let wants_readiness = matches!(
-                            tokio::time::timeout(Duration::from_millis(200), socket.read(&mut buf))
+                        // Reset backoff after a healthy accept so a past transient error doesn't
+                        // keep the delay elevated.
+                        backoff = initial_backoff;
+
+                        // Handle each connection in its own task: the request read is bounded by a
+                        // short timeout, but a client that connects and sends nothing would still
+                        // hold the accept loop for the full timeout, delaying every other probe.
+                        // A per-connection task keeps the accept loop free. `node_mode` is a cheap
+                        // watch receiver, cloned per connection.
+                        let node_mode = node_mode.clone();
+                        tokio::spawn(async move {
+                            let mut buf = [0u8; 256];
+                            let wants_readiness = matches!(
+                                tokio::time::timeout(
+                                    Duration::from_millis(200),
+                                    socket.read(&mut buf),
+                                )
                                 .await,
-                            Ok(Ok(n)) if n > 0 && request_targets_readiness(&buf[..n])
-                        );
+                                Ok(Ok(n)) if n > 0 && request_targets_readiness(&buf[..n])
+                            );
 
-                        let response = if wants_readiness {
-                            readiness_response(*node_mode.borrow())
-                        } else {
-                            // Any other path (incl. `/`) is a liveness probe: the process is up.
-                            LIVE_200
-                        };
+                            let response = if wants_readiness {
+                                readiness_response(*node_mode.borrow())
+                            } else {
+                                // Any other path (incl. `/`) is a liveness probe: process is up.
+                                LIVE_200
+                            };
 
-                        // write response, ignore errors (client disconnect, etc.), then drop.
-                        if let Err(e) = socket.write_all(response).await {
-                            tracing::error!(target: "healthcheck", ?e, "error writing healthcheck response");
-                        }
+                            // write response, ignore errors (client disconnect, etc.), then drop.
+                            if let Err(e) = socket.write_all(response).await {
+                                tracing::error!(target: "healthcheck", ?e, "error writing healthcheck response");
+                            }
+                        });
                     }
                     Err(ref e) if matches!(
                         e.kind(),
--- a/crates/middleware/orchestrator/src/types/health.rs
+++ b/crates/middleware/orchestrator/src/types/health.rs
@@ -1,65 +1,68 @@
-//! TCP health + readiness endpoint for monitoring service availability.
+//! HTTP health + readiness endpoint for monitoring service availability.
 //!
-//! A minimal HTTP/1.1 server exposing two probes:
+//! Built on `axum` (already a workspace dependency, same as `consensus-metrics`' Prometheus
+//! server) exposing two probes:
 //! - **liveness** (any path, e.g. `GET /`): always `200 OK` while the process is accepting
 //!   connections — the pre-existing behaviour.
 //! - **readiness** (`GET /readyz`, alias `GET /ready`): `200 OK` only when the node is actually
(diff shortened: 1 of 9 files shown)


crates/middleware/orchestrator/src/types/health.rs
```diff
@@ -1,73 +1,126 @@
...
-/// This service is off by default, but can be enabled through the CLI node command.
-/// Each connection is handled synchronously in the main accept loop.
-/// No connection limits or rate limiting are implemented.
-/// Connections are immediately closed after sending response.
+/// This endpoint accepts connections from any source. Node operators must protect it with a
+/// firewall. It is off by default and enabled via the CLI node command. Each connection is
+/// handled in the accept loop and closed immediately after the response.
 ///
-/// To enable on node startup, use `rayls-network node --enable-healthcheck`.
+/// To enable on node startup, use `rayls-network node --healthcheck <PORT>`.
```
procdump, line 46: As you're changing parameter names - flagging again that this might break the test env as well as deployments?

luishsr: Dug into this — the CLI flag itself is unchanged by this PR: `--healthcheck <PORT>` (`pub healthcheck: Option<u16>` in `node.rs`) already exists on `main` today, and nothing in this PR renames or retypes it. What changed here is only the doc comment on `health.rs`, which said `--enable-healthcheck` — a name that doesn't match the actual flag. I traced that mismatch back to `c08dd6e` ("Porting codebase from old repository", the original import) — it's been stale since before any of us touched this file; this PR is the first to fix the doc text to match the real flag. I also grepped `axyl-automation-testing` and `etc/` for `--enable-healthcheck`/`enable_healthcheck` — zero hits, so nothing in the test env references the old (never-actually-real) name either. No Prometheus metric names were renamed in this PR either (the new ones — `vote_request_rejections`, `committee_size`, `validator_participation` — are net-new, nothing existing renamed). So I don't think there's a live break here, but let me know if you had a specific test-env or deployment reference in mind that I should check directly — happy to look at anything specific.

crates/middleware/orchestrator/src/types/health.rs
```diff
@@ -1,73 +1,126 @@
-//! Simple TCP healthcheck endpoint for monitoring service availability.
+//! TCP health + readiness endpoint for monitoring service availability.
 //!
-//! Implements a minimal HTTP/1.1 server that responds with status 200 to all requests.
-//! This is designed for integration with GCP load balancers and similar health monitoring systems.
+//! A minimal HTTP/1.1 server exposing two probes:
+//! - **liveness** (any path, e.g. `GET /`): always `200 OK` while the process is accepting
+//!   connections — the pre-existing behaviour.
+//! - **readiness** (`GET /readyz`, alias `GET /ready`): `200 OK` only when the node is actually
+//!   participating — voting (`CvvActive`) or an operational observer — and `503 Service
```
procdump, line 8: It seems pretty odd to me to use 503 when the validator is CvvInactive. What's the reason behind this decision?

crates/middleware/orchestrator/src/types/health.rs
```diff
@@ -1,73 +1,126 @@
...
+                                .await,
+                                Ok(Ok(n)) if n > 0 && request_targets_readiness(&buf[..n])
+                            );
+
+                            let response = if wants_readiness {
+                                readiness_response(*node_mode.borrow())
+                            } else {
+                                // Any other path (incl. `/`) is a liveness probe: process is up.
+                                LIVE_200
+                            };
```
procdump: could you have used some framework for exposing this data instead of doing it the raw way string-matching the request that comes from the socket - i'd also argue that this makes the API very fragile later to extend.

bdimitrov-netzine:
The Health metric would be really useful if it can be applied to Observers. Since we send transactions to them directly it would benefit us if we can mark them as unhealthy and remove them from the existing LoadBalancer.
Because Observers are never going to be cvvActive the current implementation can not be used for them.

Right now we get alerted for an unhealthy Validator by how behind it is from the chain's tip.

crates/middleware/orchestrator/src/types/health.rs
luishsr: Explained this in detail in the module doc (see `health.rs`'s top-level comment now) — short version: liveness and readiness answer different questions on purpose. Liveness = "is the process alive, should we avoid restarting it" — a CvvInactive validator is alive and mid-recovery, restarting it would only set it back further, so liveness correctly stays 200. Readiness = "is it currently able to do its job, should traffic/expectations route to it right now" — CvvInactive means the committee member is temporarily not voting or proposing, which is exactly what readiness exists to surface. 503 specifically is the standard HTTP status for "temporarily can't serve, don't route here, but don't tear anything down" — it's what makes a load balancer pull the instance out of rotation without restarting it, and add it back automatically once the probe flips to 200 (caught up, voting again). A 200 would hide real non-participation; a 4xx/5xx-as-error would misrepresent a temporary state as a request/server failure.

crates/middleware/orchestrator/src/types/health.rs
luishsr: Done in ad9a91a — rewrote this on axum (already a workspace dependency; consensus-metrics' Prometheus server uses the exact same pattern). Routing, connection lifecycle, and accept-loop error/backoff handling are now the framework's job instead of ~50 lines of hand-rolled byte-parsing and retry logic. Adding an endpoint going forward is one `.route("/path", get(handler))` call, which was the main point of your "fragile to extend" concern.

luishsr:
@bdimitrov-netzine — good catch, and you're right that the current implementation can't help here: \`readiness_response\` treats \`mode.is_observer()\` as unconditionally ready, so a stuck/far-behind Observer reports the same 200 as a healthy one. It genuinely can't distinguish them today.

The reason I haven't rolled this into this PR: unlike the two fixes above, this needs a real design decision, not just a code change. Sketch of what it'd take:

- \`ConsensusBus::executed_anchor()\` already exposes a \`watch::Sender<ConsensusHeader>\` — the epoch/round of the last block this node executed (this is the same mechanism behind the `executed_anchor` monotonicity fix from the June-21 halt). Subscribing to it here would mirror exactly how \`node_mode\` is already threaded into \`HealthcheckServer::spawn\`.
- Readiness for an Observer would then be: has \`executed_anchor\` advanced within some threshold window? If it's been stale too long, the Observer is behind/stuck → 503.
- The open question is the threshold itself — a flat number of seconds, or something relative to the network's epoch/round cadence? That's a policy call I don't think I should make unilaterally, since it directly decides how twitchy vs. lenient your load-balancer eviction gets.

Happy to build this once we've got a number (or a way to derive one) — want to open a tracked issue for it, or would you rather we just pick a threshold here and I follow up in a separate PR?

bdimitrov-netzine:
> @bdimitrov-netzine — good catch, and you're right that the current implementation can't help here: `readiness_response` treats `mode.is_observer()` as unconditionally ready, so a stuck/far-behind Observer reports the same 200 as a healthy one. It genuinely can't distinguish them today.
> 
> The reason I haven't rolled this into this PR: unlike the two fixes above, this needs a real design decision, not just a code change. Sketch of what it'd take:
> 
>     * `ConsensusBus::executed_anchor()` already exposes a `watch::Sender` — the epoch/round of the last block this node executed (this is the same mechanism behind the `executed_anchor` monotonicity fix from the June-21 halt). Subscribing to it here would mirror exactly how `node_mode` is already threaded into `HealthcheckServer::spawn`.
> 
>     * Readiness for an Observer would then be: has `executed_anchor` advanced within some threshold window? If it's been stale too long, the Observer is behind/stuck → 503.
> 
>     * The open question is the threshold itself — a flat number of seconds, or something relative to the network's epoch/round cadence? That's a policy call I don't think I should make unilaterally, since it directly decides how twitchy vs. lenient your load-balancer eviction gets.
> 
> 
> Happy to build this once we've got a number (or a way to derive one) — want to open a tracked issue for it, or would you rather we just pick a threshold here and I follow up in a separate PR?

Thank you. 
In this case then, I think it is best to open a dedicated issue  and a separate PR.

procdump (dismissed):

procdump:
@luishsr rebase please

kstoykov (dismissed):

kstoykov (approved):

Issue #103, opened 2026-08-07
Replace fixed APY with a revenue-based reward curve for staking


lucasrparfin:
# Replace fixed APY with a revenue-based reward curve for staking

## Problem
Committing to a fixed APY is problematic — it's hard to guarantee the calculation. Downtime, backup stops, or geographic factors can all affect the result, generating community complaints and forcing us to justify the math after the fact. A fixed APY is also a liability: it doesn't scale, and with more than 2B RLS staked in the future the APY must reduce automatically.

## Proposal
A self-regulating reward curve where APY is total annual emission divided by RLS currently staked. As participation grows, the same reward pool is shared across more stakers, so yield compresses automatically — no manual intervention.

Emission has two sources:
- **Base monthly** — committed amount from the Foundation treasury; guarantees a reward floor and anchors early growth.
- **Variable monthly** — tied to network revenue (fees, institutional services, partner activity); grows as the network grows and is designed to eventually replace the base.

### Formula

    APY = ( Emission_annual ÷ RLS_staked ) × 100
    Emission_annual = ( Base_monthly + Variable_monthly ) × 12

## Emission evolution
- **Phase 1 (today → ~2y):** Foundation-backed bootstrapping. Emission = Base + Variable
- **Phase 2 (2–3y):** Base decreases, variable grows. Emission = Base(↓) + Variable(↑)
- **Phase 3 (long term):** Foundation base retired, revenue-only. Emission = Variable

## In scope
- Live APY on the staking dApp, updated every epoch
- Emission breakdown visible to users (base vs. revenue)
- Estimated yield calculator (amount → projected annual return)
- APY curve chart on the staking page and public docs
- Foundation controls to adjust base emission and variable rate via a defined approval process
- Public phase roadmap at launch

## Out of scope
- Any fixed or guaranteed APY
- Per-validator APY differentiation (uniform curve in v1)
- Automatic phase transitions (each shift requires a governance vote)
- Liquid / tokenised staking positions
- APY prediction or forecasting tools

## Reference
Interactive simulator: https://reisbr.github.io/StakingCurve/
https://parfin.slack.com/archives/C08JMNHN7EW/p1785489677647519

luishsr:
Opened #117 — a proof-of-concept implementation of the revenue-based reward curve described above.

[...]

PR #117, opened 2026-08-18
feat(fees): RewardCurve POC for revenue-based staking APY

luishsr:
## Summary

Proof-of-concept for #103 — replacing the fixed, admin-set staking APY with a self-regulating,
revenue-based reward curve.

- Adds a standalone `RewardCurve` contract (`rayls-contracts/src/fees/RewardCurve.sol` +
  `IRewardCurve.sol`) implementing #103's formula: `APY = (base + variable monthly emission) × 12
  / RLS staked`. There is no storage field for APY anywhere — only the emission inputs
  (`baseMonthlyEmission`, `variableMonthlyEmission`) are governance-settable, so a fixed or
  guaranteed APY can't be configured by construction.
- Wires the curve into `RewardDistributor`: each track's target APY is read from a configured
  `RewardCurve` on every `distributeRewards()` call — on-chain, not pushed by an off-chain
  keeper. `address(0)` disables it and falls back to the existing static bps unchanged, output is
  clamped to `MAX_APY_BPS`, and a `try/catch` means a broken curve can never block epoch close.
- 21 isolated unit tests (`RewardCurveTest.t.sol`): APY compression as stake grows, APY rising
  with recorded revenue, a phase-transition scenario that preserves APY, an overflow-guard cap
  with tests, breakdown/estimator/curve-preview correctness, access control, and a monotonicity
  fuzz test.
- 4 integration tests in `RewardDistributorExtendedTest.t.sol` driving the curve through the real
  `RewardDistributor`/`DelegationPool` distribution pipeline: curve-driven APY compressing the
  realized payout across both tracks of #85's two-tier split (Track A priority / Track B open
  tier), plus the disabled / reverting / clamped fallback paths.

## How the flow works

Agreed with @mlukanova in review:

1. Admin calls `recordRevenue(amount)` on `RewardCurve` each month.
2. Admin sets `setBaseMonthlyEmission` (once, until it needs changing).
3. Epoch ends, reward distribution is triggered, hybrid reward weights are used for the split
   as they are today.
4. The distribution flow itself is unchanged — only the **APY calculation** comes from the new
   formula. Funding is topped up automatically from `RLSAccumulator` via the existing
   `RewardDistributor._pullAccumulatorTopUp`, so the admin only ever tops up the accumulator.

Tracking/observability for the flow comes from `RewardCurve`'s events (`RevenueRecorded`,
`BaseMonthlyEmissionUpdated`, `MonthlyRevenueReset`, `PhaseTransitioned`), the recorded
validator participation, and validator/delegator stakes.

`FeeAggregator` is deliberately **not** involved: `pendingValidatorRls` is the amount already
being paid out to validators after a swap (and can include leftovers from failed swaps), not an
inflow that should fund future emission — so it is the wrong revenue signal. An earlier revision
of this PR wired it; that has been removed.

**This is a proof-of-concept, not the production rollout.** Not addressed here, and flagged in
code comments: a mismatch between `RewardCurve`'s uncapped output and
`RewardDistributor.MAX_APY_BPS` (handled by clamping for now), and a "defined approval process"
(timelock) on the governance setters — they're immediate admin transactions today, same as
`RewardDistributor`'s existing setters.

Relates to #103.

## Surface areas touched

- [ ] Execution / EVM
- [x] On-chain contracts (`rayls-contracts/`)
- [ ] Tests only

## Breaking / compatibility

None. The curve is opt-in and additive: `RewardDistributor` behaves exactly as before until
`setRewardCurve` / `setOpenTierRewardCurve` are explicitly called.

## Test plan

- [ ] `cd rayls-contracts && forge build && forge test` — full suite, 502 tests, expect 0 failures
- [ ] `forge test --match-path "test/fees/RewardCurveTest.t.sol" -vv` — the 21 isolated unit tests
- [ ] `forge test --match-test "curveDrivenApy|rewardCurve" -vv` — the real-pipeline integration tests


rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+        return _getRewardCurveStorage().baseMonthlyEmission;
+    }
+
+    /// @inheritdoc IRewardCurve
+    function variableMonthlyEmission() external view override returns (uint256) {
+        return _getRewardCurveStorage().variableMonthlyEmission;
+    }
+
+    /// @inheritdoc IRewardCurve
+    function revenueReporter() external view override returns (address) {
```
ncervaens: lets use revenue reporter as a role in AccessControl

rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+    function setBaseMonthlyEmission(uint256 newBase) external override onlyRole(DEFAULT_ADMIN_ROLE) {
+        if (newBase > MAX_MONTHLY_EMISSION) revert EmissionTooLarge();
+        RewardCurveStorage storage $ = _getRewardCurveStorage();
+        uint256 oldBase = $.baseMonthlyEmission;
+        $.baseMonthlyEmission = newBase;
+        emit BaseMonthlyEmissionUpdated(oldBase, newBase);
+    }
+
+    /// @inheritdoc IRewardCurve
+    function recordRevenue(uint256 amount) external override onlyRevenueReporter {
```
ncervaens: lets use revenue reporter as a role in AccessControl

rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+        if (amount == 0) revert ZeroAmount();
+        RewardCurveStorage storage $ = _getRewardCurveStorage();
+        uint256 newVariable = $.variableMonthlyEmission + amount;
+        if (newVariable > MAX_MONTHLY_EMISSION) revert EmissionTooLarge();
+        $.variableMonthlyEmission = newVariable;
+        emit RevenueRecorded(amount, newVariable);
+    }
+
+    /// @inheritdoc IRewardCurve
+    function resetMonthlyRevenue() external override onlyRole(DEFAULT_ADMIN_ROLE) {
```
ncervaens: only role REVENUE_REPORTER

rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+        RewardCurveStorage storage $ = _getRewardCurveStorage();
+        uint256 cleared = $.variableMonthlyEmission;
+        $.variableMonthlyEmission = 0;
+        emit MonthlyRevenueReset(cleared);
+    }
+
+    /// @inheritdoc IRewardCurve
+    function setRevenueReporter(address newReporter) external override onlyRole(DEFAULT_ADMIN_ROLE) {
```
ncervaens: only role REVENUE_REPORTER

rayls-contracts/src/interfaces/IRewardCurve.sol
```diff
@@ -0,0 +1,111 @@
...
+ *      reported network revenue. Neither is minted — both model an existing-reserve / real-revenue
+ *      payout budget, mirroring RewardDistributor + RLSAccumulator's existing non-inflationary design.
+ * @dev This contract does not custody or transfer RLS tokens — it is pure emission bookkeeping and
+ *      APY math. Real payout stays in RewardDistributor/RLSAccumulator.
+ */
+interface IRewardCurve {
+    /// @notice Emission funding phase, per issue #103's roadmap. Purely an observable marker —
+    ///         no transition logic is gated on it; phase changes are always an explicit governance
+    ///         (admin) action here, matching "each shift requires a governance vote".
+    enum Phase {
```
ncervaens, line 23: Maybe describe here each phase

rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+        variableApyBps = _apyBps($.variableMonthlyEmission * MONTHS_PER_YEAR, rlsStaked);
+    }
+
+    /// @inheritdoc IRewardCurve
+    function estimateYield(uint256 amount, uint256 rlsStaked) external view override returns (uint256) {
+        return (amount * _currentApyBps(rlsStaked)) / BPS_DENOMINATOR;
+    }
+
+    /// @inheritdoc IRewardCurve
+    function previewCurve(
```
ncervaens, line 128: apply the fix suggested by claude's review for DoS

rayls-contracts/src/fees/RewardCurve.sol
```diff
@@ -0,0 +1,204 @@
...
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(address admin_) external initializer {
+        if (admin_ == address(0)) revert ZeroAddress();
+
+        __AccessControl_init();
+        __UUPSUpgradeable_init();
```
ncervaens: use the latest OZ libraries which do not require __UUPSUpgradeable_init()

rayls-contracts/src/fees/RewardCurve.sol
luishsr: Done in 2005369 — replaced the single settable `revenueReporter` address with a `REVENUE_REPORTER_ROLE` AccessControl role, granted to `admin_` at `initialize()`. Multiple reporters can now be authorized via the inherited `grantRole`/`revokeRole`.

rayls-contracts/src/fees/RewardCurve.sol
luishsr: Same fix — see reply above.

rayls-contracts/src/fees/RewardCurve.sol
luishsr: Done — `recordRevenue` and `resetMonthlyRevenue` are both gated by `REVENUE_REPORTER_ROLE` now (paired since whoever manages the revenue-reporting cycle also closes it out each month).

rayls-contracts/src/fees/RewardCurve.sol
luishsr: Same fix — see reply above.

rayls-contracts/src/interfaces/IRewardCurve.sol
luishsr: Added doc comments for each `Phase` value describing what it represents, per issue #103's phase roadmap (FoundationHeavy/Mixed/RevenueOnly).

rayls-contracts/src/fees/RewardCurve.sol
luishsr: The overflow-DoS fix (MAX_MONTHLY_EMISSION cap on setBaseMonthlyEmission/recordRevenue) was already in this PR before your review (commit 54f09db). I've also now bounded previewCurve's own caller-supplied array length here specifically (MAX_PREVIEW_LEVELS = 200) as additional defense-in-depth, since an unbounded array is a separate (if narrower) DoS surface if this is ever called from another contract's transaction rather than an off-chain eth_call.

rayls-contracts/src/fees/RewardCurve.sol
luishsr: Confirmed — OZ v5.4.0's `__UUPSUpgradeable_init()` is an empty no-op (no state to initialize), so I dropped the call. Note the rest of the repo's existing contracts (RewardDistributor, RLSAccumulator, DelegationPool) still call it — left those untouched to keep this PR scoped, happy to clean those up repo-wide in a separate follow-up if you'd like.

mlukanova:
I think we should close this PR and proceed with the actual implementation in another. The UI is only for POC/demo purposes and is not needed in the axyl repo. `RewardCurve.sol ` is in isolation from the `RewardsDistributor` and the whole tokenomics flow. I suggest we move only the relevant code with the addition of the wiring to the exiting contracts to a new PR. Alternatively we can proceed with this and apply those changes here. @luishsr @ncervaens @lucasrsilva

crates/middleware/rewards/examples/tally_replay.rs
mlukanova, line 1: We can include the tally_replay and tally_sweep as part of the cli db tool implementation. https://github.com/raylsnetwork/axyl/pull/148 Seems more appropriate place.

rayls-contracts/playground/README.md
mlukanova, line 1: The UI playground should not part of the axyl repo. It's only purpose was manual interaction with the POC state of the RewardCurve contract. Let's keep the axyl repo clean.

rayls-contracts/src/fees/FeeAggregator.sol
mlukanova, line 1: The FeeAggregator's idea is to gather two revenue streams:
1. The collected fees from network transactions
2. Direct payments in stables from PNs or other customers - this is unused so far
The goal of the integration was to use some of the functionality for the reward curve from my point of view. 

There is one call to `RewardCurve.recordRevenue` setting the `pendingValidatorRls` as revenue. It seems like the wrong parameter.  The pendingValidatorRls can also be some leftovers from failed swaps. Let's see if this is the best way to set it.
`baseMonthlyEmission` ("Foundation commitment") has no on-chain accounting mechanism - it's a bare onlyRole(DEFAULT_ADMIN_ROLE) value; `RewardCurve` explicitly doesn't custody RLS so we would need a strategy to get those amounts of tokens from somewhere. 
`RLSAccumulator` is topped up independently via `deposit()` - we can use it as a "Foundation commitment" or Revenue stream but it needs some restrictions - how much is allowed to take for a month 

Is your idea different? Where is the revenue and Foundation commitment tokens being taken from?

luishsr: Agreed on all points, and thanks for spelling out the intended flow — that resolved the open question. Removed the FeeAggregator wire in 6eab788.

You're right that `pendingValidatorRls` is the wrong parameter, and for a second reason too: it's additive across epochs (the bucket only clears on a *successful* transfer), so a retried epoch would have reported a multi-epoch backlog as one month's revenue into an accumulating `variableMonthlyEmission` — double-counting on top of it being an outflow rather than an inflow.

So the PR now matches your flow exactly, and needed no new plumbing to get there:

1. Admin calls `recordRevenue(amount)` monthly on `RewardCurve` (gated by `REVENUE_REPORTER_ROLE`).
2. Admin sets `setBaseMonthlyEmission` once, until it needs changing.
3. Epoch ends, distribution triggers, hybrid weights are used as today.
4. Flow unchanged — only the APY calculation comes from the new formula, and `RewardDistributor._pullAccumulatorTopUp` already draws from `RLSAccumulator` to cover the shortfall against the curve-derived target. So the admin only ever tops up the accumulator.

The `RewardDistributor` -> `RewardCurve` APY wire is kept, since that's the half that actually implements the new formula. Removal was surgical: `FeeAggregator.sol`, `IFeeAggregator.sol` and `FeeAggregatorTest.t.sol` are now byte-identical to their pre-wiring state.

On your point about restricting how much can be taken per month from `RLSAccumulator` — that isn't in this PR. The curve bounds it indirectly (the top-up only ever covers the shortfall against the curve-derived target, so emission is capped by the formula rather than by the accumulator balance), but there's no explicit per-month withdrawal cap on the accumulator itself. Happy to add one here or track it for the rollout issue (#79) — let me know which you'd prefer.

502/502 tests pass.

rayls-contracts/playground/README.md
luishsr: Agreed — removed in 78fabb2. The playground served its purpose during review and doesn't belong in the repo; it's recoverable from d0fc7fd if we ever want it as a standalone repo. Nothing outside that directory referenced it.

crates/middleware/rewards/examples/tally_replay.rs
luishsr: Agreed — dropped both from this PR in 04b13cb, so they can land with the DB inspection CLI instead. That also removed the `serde_yaml` dev-dependency they were the only users of, returning `Cargo.toml`/`Cargo.lock` to their pre-64c264a state.

I haven't pushed anything to #148 since it's @ricardopinto's branch — the tools are recoverable from 64c264a whenever it makes sense to fold them in. Worth noting I have a local extension to `tally_sweep` (a historical entitlement average that replays `applyIncentives`' 80/10/10 weight blend across the full epoch history) that I kept aside; happy to hand that over too if it's useful there.

mlukanova (approved):

procdump (approved):

PR #121, opened 2026-08-20
feat: adopt reth pool maintenance and persist the txpool across restarts

bronxyz:
## Summary

- Replace the hand-rolled canonical-state pool update with reth's `maintain_transaction_pool_future`, and hold locally submitted transactions to the same fee and eviction rules as external ones (`no_local_exemptions`). In-flight marks release on a dedicated canonical-stream task, with the TTL sweep and membership reconcile driven from the 30s engine gap-check tick so a burst cannot starve release.
- Delete the bypass validator and the orphan-batch re-introduction: sealed-but-uncommitted transactions now stay pending and marked in flight, so nothing is re-collected from `NodeBatchesCache` and bypass-validated back in.
- Persist every worker pool (pending + queued transactions and their marks) on graceful shutdown and reload it on boot, so sealed-but-uncommitted transactions, which no peer can re-supply, survive a restart. A safety-net snapshot runs before the unbounded engine-drain await; both writes are tmp-then-rename so the post-drain snapshot wins on the healthy path.

Stack 2/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None on the wire. Behavior change: local transactions lose their fee/eviction exemption, and a node now writes a txpool backup file under its datadir on graceful shutdown. A transaction executed during the drain is rejected nonce-too-low on reload and its mark reconciled away.

## Test plan

- [x] `atomic_write_replaces_existing_file`, `streaming_reader_accepts_legacy_json_array` in `txn_pool/backup.rs`.
- [x] Builder integration tests (`tests/it/build_batches.rs`) pass against the reth maintenance task.


crates/execution/evm/src/txn_pool/backup.rs
```diff
@@ -0,0 +1,388 @@
...
+        let queued = self.queued_transactions();
+        let total = pending.len() + queued.len();
+        let path = self.backup_path.as_ref();
+
+        if total == 0 {
+            remove_backup_file(path);
+            return 0;
+        }
+
+        let write_result = write_atomically(path, |writer| {
```
procdump, line 60: do we need to persist queued/pending txs upon graceful restarts - if we delay the start up of the stopped validator could this lead to double spending problem because the user erroneously retried with a new nonce? - also in the case of being an observer its snapshot may be used for spinning a new one inheriting the persisted txs - could this lead to an issue?

bronxyz: The txpool and evm won't allow for double spending to occur.
1. When we restart - we are going to enter the `CvvInactive` state, which doesn't produce any batches. Any tx that has been part of a block will be dropped by the pool maintenance. All remaining ones are still eligible to be included in batches when they become pending.
2. The pool snapshot must be used for local purposes only, so snapshots must not include this local pool backup for restoring purposes.

kstoykov (approved):

PR #122, opened 2026-08-20
feat: seal integrity and batch-seq normalization under a new hardfork

bronxyz:
## Summary

- Never strand a sealed batch's transactions: an unacknowledged `report_own_batch` used to record an uncommittable batch as sequenced (seq consumed for a digest no header carries, successors parked until the boundary drain). The seal now surfaces it as a failure so the builder keeps the seq and the transactions selectable, and every seal failure reports a typed `SealFailureReason` to the tracker. A batch executed without its nonce-too-high transactions now frees their marks through an `UnsettledExecution` witness, so they do not sit unselectable for the full TTL.
- One node-scoped `InFlightTracker` is shared between the worker pool and each epoch's engine (`init_txn_pool_with_in_flight`), so the release path is live from the first epoch after boot.
- Add the `OutputSeqNormalization` hardfork (`0x0e`; `Never` on every real network, block 0 on local) gating three replicated-execution changes: parked batches are discarded whole at the epoch boundary instead of force-executed out of order, intra-output seq inversions are repaired in place, and the parked entries an overflow-forced jump abandons are pruned so ordering comes back instead of force-executing every later gap.

Stack 3/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

Introduces the `OutputSeqNormalization` hardfork. It is `ForkCondition::Never` on every real network until SRE schedules an activation block; all three gated behaviors change replicated output, so old and new nodes must switch at the same block. Rolling upgrade is safe while the fork stays `Never`. No wire-format changes.

## Test plan

- [x] `test_failed_report_leaves_txs_selectable_and_seq_unconsumed` (via `MockWorkerToPrimaryError`) and `test_dropped_txs_release_in_flight_marks`, both red against the pre-fix paths.
- [x] Red-first pins for each fork-gated behavior: 3 blocks pre-fix vs 2 for the boundary discard, park-and-gap-fill vs in-place repair as the rolling-safety pair, 32 -> 0 parked plus the pre-fork retention pin for the overflow prune.


kstoykov (dismissed):

procdump:

PR #123, opened 2026-08-20
fix: harden batch ordering, proposer requeue, and boot recovery

bronxyz:
## Summary

- Harden batch-ordering state: `drain_epoch` can no longer rewind the epoch on a crash-replay of an already-passed output; the boot history walk opts out of the long-read cap and fails loud instead of silently booting with empty watermarks; parked batches persist by digest (`ParkedRef`) with the snapshot taken under the lock and written outside it, and the authority map becomes a `BTreeMap` for deterministic persisted bytes. Legacy by-value blobs still decode via a fallback.
- Requeue evicted and lag-stranded proposer digests: GC keyed its horizon on `self.round` and dropped the evicted headers' digests, permanently gapping the per-authority seq stream on peers; the horizon now keys on the monotonic committed round and eviction requeues FIFO-ahead. The foreign-commit fallback retransmitted every proposed header; it now splits at a grace horizon below the commit round so ordinary lag never triggers it.
- Drop the dead `NodeBatchesCache` writes (the orphan rescue was their only reader) and the boundary flag that justified them; close three robustness gaps (a clean external shutdown misclassified as a crash inside `run_epoch`, a multi-worker in-flight wipe fenced by `ensure!`, a `debug_assert` promoted to a release assert); restore the observability trio (`ForceDrained` tracker stage, `AlreadyImported` demoted to debug, `txpool` RPC module enabled).

Stack 4/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None on the wire. The persisted batch-ordering blob changes shape; a legacy by-value blob is still decoded (pinned by `stored_and_legacy_ordering_blobs_are_distinguishable_by_decode`). `NodeBatchesCache` stays declared (stable column-family set) and is still cleared on foreign-DB sanitization. The `txpool_*` RPC module becomes available.

## Test plan

- [x] `drain_epoch_never_rewinds_to_an_already_passed_epoch`, `stored_and_legacy_ordering_blobs_are_distinguishable_by_decode`.
- [x] Red-proven proposer tests: `fallback_requeue_forgives_ordinary_commit_lag` fails at grace 0, `digests_survive_gc_advance_and_later_round_commit` fails when the requeue drops; `push_digest_never_drops` and `fallback_requeue_stops_at_the_commit_round` pin the contracts.
- [x] The el-to-cl seal test asserts no `NodeBatchesCache` row is written.
- [ ] The `run_epoch` shutdown race and the multi-worker wipe need multi-task fault injection to demonstrate red and land as hardening without failing-first tests.

### DIFF
--- a/crates/consensus/primary/src/consensus/state.rs
+++ b/crates/consensus/primary/src/consensus/state.rs
@@ -549,17 +549,9 @@ impl<DB: Database> Consensus<DB> {
             Err(e) => return Err(e),
         };
         if self.active {
-            // We extract a list of headers from this specific validator that
-            // have been agreed upon, and signal this back to the narwhal sub-system
-            // to be used to re-send batches that have not made it to a commit.
+            // Committed certificates are reported back to the primary so the proposer can drop
+            // its own committed headers and re-propose the batches of those that missed commit.
             let mut committed_certificates = Vec::new();
-
-            // Each cert is tagged with whether its subdag reaches the epoch boundary: the
-            // subscriber drops those post-boundary outputs, so the proposer must keep
-            // (not clean) their batches for rescue. Computed here where both the subdag
-            // commit_timestamp and the epoch_boundary are known; carried per-cert to
-            // the proposer (no shared transition flag).
-            let epoch_boundary = self.consensus_config.epoch_boundary();
 
             // Output the sequence in the right order.
             let csd_len = committed_sub_dags.len();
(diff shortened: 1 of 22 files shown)


kstoykov (approved):

procdump:

PR #124, opened 2026-08-20
feat: drive the batch builder through a typestate seal-ahead pipeline

bronxyz:
## Summary

- `BatchOrdering` publishes this authority's accepted batch sequence as a watch, seeded from the persisted `last_executed_seq` and advanced on in-order accept, overflow-forced accept, and parked-batch drain. It marks acceptance into execution order, not block finalization.
- Replace the poll-based `BatchBuilder` future with an async run loop over a `BatchPipeline` typestate, so an illegal build/seal ordering cannot be represented. The builder seals up to `MAX_SEAL_AHEAD` batches ahead of its own watermark and then stalls until execution advances, keeping the in-flight prefix inside the per-authority parking budget; the budget collapses to one batch inside the boundary quiesce window and the builder ends once the canonical tip reaches the epoch boundary.
- On restart the sequence resumes from the execution watermark rather than from the persisted counter alone.

Stack 5/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None. Node-local pacing only; no wire, storage, or fork changes. A fast proposer now seals at most `MAX_SEAL_AHEAD` batches past its own executed sequence.

## Test plan

- [x] `tests/it/pipeline_typestate.rs` covers the phase transitions and the seal-ahead / quiesce / closed budget.
- [x] `tests/it/build_batches.rs` runs the builder through the new `run()` loop end to end.

### DIFF
--- a/crates/consensus/worker/src/batch-builder/src/batch.rs
+++ b/crates/consensus/worker/src/batch-builder/src/batch.rs
@@ -13,25 +13,40 @@ use rayls_infrastructure_types::{
 };
 use tracing::debug;
 
-/// The output from building the next block.
+/// The transactions selected into a sealed batch, to be marked in flight on quorum.
 ///
-/// Contains information needed to update the transaction pool.
+/// `#[must_use]` so a build's selection cannot be silently dropped: the caller either marks it in
+/// flight on quorum or explicitly discards it.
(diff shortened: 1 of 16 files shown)


kstoykov (approved):

procdump:

PR #125, opened 2026-08-20
feat: sender-affinity load balancing and direct observer forwarding

bronxyz:
## Summary

- Add the `SenderAffinityLoadBalancing` hardfork (`0x0f`; `Never` on real networks, block 0 on local). Under the fork the committee-slot digest keys on the recovered sender, so one validator owns a sender's whole nonce chain instead of consecutive ranges scattering across pools and parking nonce-gapped. `CommitteeSlots` adds a ring-walk `covers()` predicate so a down owner's senders fail over to the next live slot, agreed across nodes by the sorted committee order.
- Forward an observer's transactions directly to the committee: a per-epoch forwarder submits pending transactions by request-response to the validator owning each sender's slot (gossip before the fork), gates re-sends behind the catch-up watermark while first sends still flow, and ring-walks to the next live validator on failover. The ack carries the hashes the owner rejected as nonce-too-low, which are suppressed as acked-stale. An observer no longer runs a batch builder.
- Stickiness in Observer mode now keys on the operator's `--observer` flag rather than on the node's current mode, so a staked validator that booted into Observer to catch up can be promoted again; an empty slot table mid-transition drops the message instead of reaching the modulo (division by zero aborts the node).

Stack 6/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

- Wire: `WorkerRequest::SubmitTxns` is appended last and `WorkerResponse::SubmitTxns` after `Error`, so every existing variant keeps its bcs index; an un-upgraded peer answers the new request with `Error` and the sender retries the next validator. `WorkerGossip::Txn` moves to `Vec<Bytes>`, which bcs-encodes identically to `Vec<Vec<u8>>`.
- Hardfork: `SenderAffinityLoadBalancing` is `Never` on every real network until an activation block is scheduled. The fork gate reads the local canonical tip, so validators can briefly disagree across the fork block; the worst case is a duplicate include that fails nonce-too-low at execution, never a consensus fork.
- Behavior: observers stop producing batches and forward instead. Inbound submit fan-out is bounded by a semaphore.

## Test plan

- [x] bcs-index pins: `Error` stays variant index 3 on `WorkerResponse` and round-trips as itself; `SubmitTxns` lands at index 4 (response) and 3 (request).
- [x] `save_mark_backup` / `load_mark_backup` round trip for forwarding marks, and the pin that a sealing snapshot writes no file.
- [x] Forwarder unit tests: `is_caught_up_gates_on_local_lag`, `should_send_flows_first_sends_but_gates_resends`, `validate_stale_keeps_only_hashes_this_node_sent`.
- [ ] The observer-stickiness and zero-slot guards need a mid-transition committee fault to demonstrate red and land as hardening without failing-first tests.


crates/consensus/worker/src/batch-validator/src/validator.rs
```diff
@@ -118,25 +109,38 @@ impl BatchValidation for BatchValidator {
...
+                }
+                let owner = self.slot_digest(tx) % slots.size();
+                // Under sender-affinity a down owner's senders fail over to the next live slot on
+                // the ring; pre-fork ownership is exact, with no failover.
+                let mine = if self.sender_affinity_active() {
+                    slots.covers(owner)
+                } else {
+                    owner == slots.own_slot
+                };
+                if !mine {
```
procdump, line 140: thinking over this affinity instead of gossip approach - here's the following resistance concern - what if a naughty owner validator dishonestly decides to skip txns?

kstoykov (approved):

procdump:

PR #127, opened 2026-08-20
perf: hot-path allocation, hashing, and backpressure optimizations

bronxyz:
## Summary

- Cut allocation and hashing on the hot paths: `Batch.transactions` carries `Bytes` so a batch cloned through seal, quorum, and ordering refcounts its payloads instead of deep-copying up to 2MB; the consensus-cert, executed-batch, and pending-parent maps key on the digest byte hasher instead of SipHash; certificates are sized with `bcs::serialized_size` instead of being encoded twice; batch-existence checks probe the key without decoding the value; `validate_batch` takes `&SealedBatch`.
- Cut per-peer overhead: the quorum waiter awaits each peer's ack directly in its `FuturesUnordered` instead of spawning a task per peer; the certifier, fetcher, and worker seal loops reuse a pinned timer and drop per-iteration allocations; kad record puts are budgeted to ten per source per rolling minute and deduplicated before the signature verify.
- Keep backpressure honest: the proposer throttles only proposing (digests, parents, and ticks keep draining) and reads the suspended-certificate count from the certificate manager's typed watch rather than a metrics gauge; the engine stops admitting consensus output at queue capacity so execution lag propagates back to consensus instead of being absorbed as memory.

Stack 8/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None on the wire: `Vec<Bytes>` and `Vec<Vec<u8>>` bcs-encode identically (ULEB128 length + raw bytes), so the batch encoding and every stored digest stay byte-identical; the container is a node-local choice. Behavior: a peer exceeding ten kad puts per minute has the excess dropped before verification; the recently-validated dedup set is bounded to a one-minute window.

## Test plan

- [x] `batch_encoding_golden_bytes` pins the `Batch` wire encoding; `tx_payload_bcs_identical_for_vec_u8_and_bytes` cross-decodes the two containers.
- [x] `test_kad_put_budget_drops_flood`, `test_kad_put_stale_duplicate_deduped_before_signature_verify`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7792,6 +7792,7 @@ dependencies = [
  "assert_matches",
  "async-trait",
  "backoff",
+ "bcs",
  "consensus-metrics",
  "eyre",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7849,7 +7849,6 @@ version = "1.3.0"
 dependencies = [
  "assert_matches",
  "async-trait",
- "consensus-metrics",
  "eyre",
  "futures",
  "indexmap 2.13.0",
--- a/bin/rayls-replay/src/replay.rs
+++ b/bin/rayls-replay/src/replay.rs
@@ -6,7 +6,7 @@ use crate::{
     plan::PlanBlock,
 };
 use rayls_execution_evm::{reth_env::RethEnv, ExecutedBlock};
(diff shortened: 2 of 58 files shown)


kstoykov (approved):

procdump:

PR #131, opened 2026-08-20
feat: added multi arch builds support

bdimitrov-netzine:
## Summary

- Adding build support for arm64

Closes #

## Surface areas touched

- [ ] Execution / EVM
- [X] CI / build (`.github/workflows/`, `Makefile`)
- [ ] Tests only


kstoykov:
Can we address fail-fast commend from claude as it could be useful in future.

procdump:
`arm64` is a RISC arch and hence threads may observe instructions reordered if incorrectly synchronized. `amd64` however is a CISC arch and such reorderings are less likely. At any rate severe testing on `amd64` **doesn't** guarantee that the same code will work flawlessly on `arm64` - just flagging this distinction to always have in mind.

PR #126, opened 2026-08-20
refactor: extract the engine-update loop with an injectable interval

bronxyz:
## Summary

- Move the node-scoped engine-update task body into `engine_update_loop`, leaving `spawn_engine_update_task` a thin spawner, and take the gap-check interval as a parameter (still 30s in production) so a node harness can drive one maintenance tick without the cadence.

Pure extraction: the recently-executed push, gap check, and per-pool sweep/reconcile are unchanged. Stack 7/9 of the txpool in-flight tracker and observer-forwarder series.

## Surface areas touched

- [ ] Consensus protocol (primary / worker / network / state-sync)
- [ ] Execution / EVM
- [ ] JSON-RPC (`eth_*`, `rayls_*`, faucet)
- [x] Middleware (orchestrator / processor / bridge)
- [ ] Infrastructure (types / storage / config / network-cli)
- [ ] On-chain contracts (`rayls-contracts/`)
- [ ] Operations (`etc/`, scripts, Docker, compose)
- [ ] CI / build (`.github/workflows/`, `Makefile`)
- [ ] Documentation only (`doc/`, in-crate READMEs, root docs)
- [ ] Tests only

## Breaking / compatibility

None.

## Test plan

- [x] Behavior-preserving refactor; `make check` on the stack tip and CI on this branch.

PR #135, opened 2026-08-26
fix: stack review follow-ups for #120-#127

bronxyz:
## Summary

- Fix the findings from the review-bot passes across the stack, verified against the code at the stack tip. One commit per finding; each commit body names the PR it addresses (#120, #121, #123, #124, #125 x2, #127).
- The #127 livelock finding turned out to be a dead brake, not a sticky one: the suspended-cert watch has no receiver and the proposer borrows the sender, so every `send` failed and stored nothing - the count read 0 forever. `send_replace` on both suspend and drain.
- The remaining fixes: unchecked `MARK_BACKUP_VERSION` on restore plus `#[must_use]` on the arm handles; infallible `decode` on a parked body at boot; `ok_or_eyre` literal braces; `elapsed_ms` stamped after the quorum result; a bcs pin for `WorkerGossip::Txn(Vec<Bytes>)`; a swallowed `JoinError` in the pre-drain txpool backup.

Stack 9/9 of the txpool in-flight tracker and observer-forwarder series.

Review findings deliberately not taken, with the reason:

- `max_tx_lifetime: 5 min` (#121): the review misread the knob; reth applies it to `queued_transactions()` only (`maintain.rs:276-284`), a pending tx is never lifetime-evicted. Documented at the site.
- `biased;` on the engine-update `select!` (#126): arms commute, a missed shutdown pick costs one idempotent tick;
- 300 ms sleep in `test_failed_report_leaves_txs_selectable_and_seq_unconsumed` (#122): it gives a buggy seal path room to mark/persist; the correct path never does either, so there is nothing to poll for.
- `RethRpcModule::Txpool` (#123): operator `--http.api` config, enabled deliberately by `e460f8fd`; mempool contents are gossiped, so the privacy premise does not hold.
- Already fixed higher in the stack: `release_dropped` uncalled, `update_canonical_state_async` unwired, stale `orphan_batches` comments, redundant `gc_round == 0` guard.
- Nits rejected as noise: `connected.contains` on a 4-7 member committee, `gauge.sub(0)`, `SeqCst` in a test mock, `stash_restore` overwrite guard, `anchor: 0` in `SealMarks::mark` (the sealing policy has `min_anchor_advance: 0`).

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None. The gossip pin proves `Vec<Bytes>` and `Vec<Vec<u8>>` bcs-encode identically; a mark backup from another schema version is now discarded on restore instead of applied, which only affects a node-local file.

## Test plan

- [x] Red-first tests for the four behavior changes: `suspended_cert_count_follows_the_drain`, `arm_rejects_a_backup_from_another_schema_version`, `reconstruct_drops_a_parked_batch_whose_body_is_corrupt`, `worker_gossip_txn_payload_is_bcs_identical_to_vec_vec_u8`.
- [x] `make check` clean; 534/534 tests in the touched crates (`primary`, `evm`, `storage`, `worker`, `batch-builder`, `orchestrator`); clippy warnings only in untouched files.


kstoykov (dismissed):

procdump:

PR #148, opened 2026-09-08
feat: Implemented DB inspection CLI

### LINKED ISSUE #144: rayls-db-inspect tool
# Build `rayls-db-inspect` — a consensus-db inspection CLI for operators

## The problem
When an epoch boundary fails to durably certify (an epoch record orphaned
network-wide — never certified/persisted on any node), a node can't restart or
sync and we're blind: there's no way to ask "what does each node actually have
on disk for epoch N?" without ad-hoc code. We hit exactly this during the
epoch-record durability incident and had to guess. We need a small, supported
CLI that reads a node's `consensus-db` offline and reports epoch-record / cert /
header state, so operators can diagnose these halts directly.

## What to build
A CLI that opens a node's `consensus-db` and inspects it. It should accept
several nodes' DBs in one run so their state can be compared side by side.

Subcommands:
- `epoch <N>` — is epoch N's record on disk, and is its cert present? A verbose
  flag dumps the record's fields. Across multiple nodes, print a network-wide
  verdict (incl. "missing on every node").
- `epochs <FROM> <TO>` — scan a range and show record / record+cert / missing per node.
- `cert <N>` — dump/compare the leader certificate at consensus header number N.
- `header <N>` — dump the consensus header at number N.
-  suggest more here so that eventually they can turn convenient to the whole debugging process - also think of recursive solutions where we need to gather info for chained data structures.

## Acceptance criteria
- [x] The four subcommands above, comparing multiple nodes in one run.
- [x] Inspect a **live** node safely (read-only), or fail clearly and document
      the stopped-node / copy workflow if that isn't possible.
- [x] Tests covering each subcommand against a fixture DB.
- [x] A short README and complete `--help` text.

ricardopinto:
## Summary

- New binary `bin/rayls-db-inspect`. Opens one or more nodes' `consensus-db` read-only and compares what they hold: epoch records, certificates, consensus headers, leader certificates.
- Safe on running nodes: nothing is written, no lock is taken. Verified against the local 4+1 network under a 5 tx/s load.
- Seven subcommands, text or `--json` output, one machine-parseable verdict line, exit status 0/1/2.
- Storage crate: `MdbxDatabase::open_read_only`, `has_table`, `table_entry_counts`, `recover`. Additive; nothing in the node calls them.
- 62 tests against seeded MDBX fixtures, including a cross-process live-writer test.

Closes #144 .

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None.

## Test plan

- [x] Check the bin/rayls-db-inspect/README.md


ricardopinto:
Consider `--back` without upper bound is not an issue. The user should be able to walk the entire chain as needed.

Regarding `compact_in_place`, left as is on purpose. The read-write open of the source doubles as recovery for a datafile left by a killed node and keeps the source opened with the same `MdbxConfig` the node uses. The verification re-open of the copy is now read-only, which removes the misleading corruption log.

procdump (changes requested):
```
rayls-db-inspect epoch                                                                                                                          
error: the following required arguments were not provided:
  --db <[LABEL=]DATADIR>
  <EPOCH>

Usage: rayls-db-inspect epoch --db <[LABEL=]DATADIR> <EPOCH>
```

I think it's far easier to have `--db` param before instead of after the commands.
Also please think of two more commands that may get pretty handy:
* get-batch
* get-tx

ricardopinto:
Regarding `get-batch` and `get-tx`, these sound like good ideas and worth following-up on separate issues.

ricardopinto:
@procdump All changes have been implemented. The  `--db` param can now be used before and/or after the commands. The `get-batch` and `get-tx` commands have been implemented too.

procdump (approved):

kstoykov (approved):

PR #154, opened 2026-09-09
feat: config flag

kstoykov:
## Summary

- Externalize chain-id and hardfork schedule from the binary into per-client YAML config files (`--config-file` / `--subnet`), replacing the baked-in `--network` hardfork profile system.
- Add `RaylsNetwork::chain_id()` and verify the datadir's genesis chain-id against the expected value at node boot, refusing to start on mismatch.
- Rename `rayls_hardforks()` to `add_rayls_hardforks_by_type()` and add `add_rayls_hardforks_by_schedule()` for external config support.
- Update mainnet chain-id from `487` to `72957` and devnet chain-id from `2017` to `503`.
- Ship `etc/test-network/config.yaml` with all four networks mirroring the baked-in schedules, plus a startup script using the config file.

Closes #

## Surface areas touched

- [x] Execution / EVM
- [x] Tests only

## Breaking / compatibility

The `rayls_hardforks()` builder method was renamed to `add_rayls_hardforks_by_type()`. Mainnet chain-id changed from `487` to `72957`. Nodes started with `--config-file` cannot also use `--network`.

## Test plan

- [x] `cargo test -p rayls-execution-evm network_profile` — 11/11 pass (ForkActivation serialization, profile parsing, schedule resolution, example yaml)
- [x] `cargo test -p rayls-execution-evm chainspec` — 15/15 pass (hardfork activation across networks)
- [x] `cargo test -p rayls-execution-evm active_profile` — 1/1 pass (external schedule reaches chain spec)
- [x] `cargo test -p rayls-network-cli` — 16/16 pass (chain-id verification, CLI parsing)
- [x] `cargo test -p rayls-infrastructure-types chain_id` — 1/1 pass (chain-id constants)
- [x] `cargo test -p rayls-batch-validator` — 25/25 pass (batch validation with updated builder)
- [x] `cargo fmt --all --check` — passes


crates/infrastructure/network-cli/src/node.rs
```diff
@@ -18,14 +20,18 @@ use rayls_infrastructure_types::Committee;
...
+};
 use tracing::*;
 
-/// Chain-ids that must never be paired with `--dev`. Mainnet only — testnet and
-/// the devnet default both use chain-id 2017, so this list cannot exclude testnet
-/// without also blocking valid local devnet use.
+/// Chain-ids that must never be paired with `--dev`. Mainnet only (`72957`) —
+/// testnet, devnet and local have distinct, non-production chain-ids.
 #[cfg(feature = "dev-single-node-setup")]
-const PROD_CHAIN_IDS: &[u64] = &[487];
```
procdump, line 28: in single node setup why changing this one?

crates/infrastructure/types/src/network.rs
```diff
@@ -32,3 +32,33 @@ impl std::fmt::Display for RaylsNetwork {
...
+    /// Mainnet is `72957`, testnet `7295799`, devnet `503` and local `487`.
+    /// The node verifies the datadir's genesis chain-id against this at boot,
+    /// so a datadir from the wrong network (or client) is refused before
+    /// anything runs.
+    pub const fn chain_id(self) -> u64 {
+        match self {
+            Self::Mainnet => 72957,
+            Self::Testnet => 7295799,
+            Self::Devnet => 503,
+            Self::Local => 487,
```
procdump, line 48: Are these different from the ones currently used now - I mean - won't introducing some id changes (if such) break existing setups?

crates/testing/e2e-tests/Cargo.toml
```diff
@@ -32,9 +32,10 @@ futures = { workspace = true, optional = true }
 escargot = { workspace = true }
 
 [dev-dependencies]
-reth = { workspace = true }
-tempfile = { workspace = true }
-serde_json = { workspace = true }
+    reth = { workspace = true }
+    tempfile = { workspace = true }
+    serde_json = { workspace = true }
+    serde_yaml = { workspace = true }
```
procdump: indent

crates/infrastructure/network-cli/src/node.rs
kstoykov: In order to match our real chain id. Basically - simplicity.

crates/infrastructure/types/src/network.rs
kstoykov: These are production IDs that we use.

crates/testing/e2e-tests/Cargo.toml
kstoykov: fixed

procdump (approved):

PR #168, opened 2026-09-14
fix/chain id consistency

kstoykov:
## Summary

- Made `Parameters.network` optional (`Option<RaylsNetwork>`); `None` means the datadir was built externally and the boot-time chain id must be supplied via `--config-file`/`--subnet` or `--network`. The node, reth env init, and epoch manager all resolve the expected chain id and hardfork schedule with the same precedence: external profile > `--network` > `parameters.network` > external-without-schedule (bail).
- Aligned dev mode to the local network (chain-id 487): the dev genesis ceremony now uses `RaylsNetwork::Local.chain_id()`, bootstrap stamps `network: local` into `parameters.yaml`, and the dev node gets `--storage.v2` enabled by default.
- Pinned e2e and chaos test suites to the local network (487) via `TEST_NETWORK` constants (ceremony `--chain-id`, node `--network`, tx spam faults).
- Removed the faucet's `--chain-id` flag; the faucet now derives the signing chain id from `reth_env.chainspec().chain_id()`.
- Retired the orphan 2017 chain-id in all fixtures (`test_genesis`, `testnet-genesis.yaml`, `create_explicit_eip1559` default, datadir-path test).

Closes #

## Surface areas touched

- [x] Execution / EVM
- [x] Operations (`etc/`, scripts, Docker, compose)
- [x] Tests only

## Breaking / compatibility

`Parameters.network` is now `Option<RaylsNetwork>`. A datadir whose `parameters.yaml` has `network: null` (or omits the key) is "external": the node must be started with a `--config-file`/`--subnet` schedule or an explicit `--network` whose chain-id matches the genesis, and it fails loudly at boot if neither is given.

Blast radius:
- **Ceremony-created datadirs are unaffected.** The old field was a non-Option `RaylsNetwork` (default `testnet`) and was always serialized, so any `parameters.yaml` written by the genesis ceremony already carries an explicit `network: <profile>` and boots as before.
- **Only hand-written `parameters.yaml` files that omit the `network:` key change behavior** — they previously defaulted to `testnet`, now resolve to "external". Such a node will not silently change chains; it refuses to start with an explicit error telling the operator to pass `--network <profile>` or a `--config-file` schedule.

No serde migration shim is added (keeping "no profile = external" unambiguous); the failure mode is a loud, actionable boot-time error rather than a silent default.

## Test plan

- `cargo check --workspace --all-targets` — passed
- `cargo +nightly-2026-06-24 fmt --all --check` — clean (the formatter CI runs)
- CI gate 1: `cargo test --workspace --exclude rayls-execution-faucet --exclude e2e-tests -- --quiet --test-threads 1` — 58 test binaries, EXIT=0
- CI gate 2: `cargo test -p rayls-infrastructure-types -p rayls-consensus-primary -p rayls-middleware-orchestrator --features dev-single-node-setup -- --quiet --test-threads 1` — EXIT=0
- CI gate 3: `cargo test -p rayls-execution-evm --features archive-replay -- --quiet --test-threads 1` — EXIT=0
- e2e default suite: `cargo test -p e2e-tests` — 4 passed, 7 ignored
- e2e restarts: `cargo test -p e2e-tests -- --ignored --test-threads 1 test_restartstt` — 1 passed (82s)
- e2e dev: `cargo test -p e2e-tests --features dev-single-node-setup -- --ignored --test-threads 1 dev_bootstraps_produces_blocks_and_includes_tx` — 1 passed (post `--storage.v2` fix). Note: `--features dev-single-node-setup` is required, since `tests/it/dev.rs` is only compiled under that feature.
- chaos: `cargo test -p chaos-framework -- --ignored --exact test_spam_wrong_chain_id` — 1 passed (40s)
- fuzz: `cargo check --all-targets` in `crates/testing/fuzz-targets/fuzz` — 225 crates, EXIT=0


procdump:
Since the resolved hardfork schedule is never persisted, a restart with an edited file (e.g. Tokenomics: 1000 -> never after block 5000) boots cleanly and continues under a schedule that contradicts the existing state, so shouldn't we write the schedule into the datadir on boot and refuse to start if any already-activated fork has changed?

kstoykov:
> Since the resolved hardfork schedule is never persisted, a restart with an edited file (e.g. Tokenomics: 1000 -> never after block 5000) boots cleanly and continues under a schedule that contradicts the existing state, so shouldn't we write the schedule into the datadir on boot and refuse to start if any already-activated fork has changed?

I agree. Let's make it as a new feature/PR as this one is for fixing chaid-id related tests.

crates/middleware/orchestrator/src/epoch_manager/worker.rs
```diff
@@ -207,13 +207,21 @@ where
...
+        // final arm is unreachable at runtime. If a future refactor ever lets a node reach
+        // it, `debug_assert!` surfaces that loudly in tests instead of silently applying an
+        // all-`Never` schedule (which would leave EIP-1559 permanently inactive).
         let hardforks = match active_profile() {
             Some(profile) => RaylsChainHardforks::new(profile.schedule()),
             None => match network {
                 Some(network) => RaylsChainHardforks::for_network(network),
-                None => RaylsChainHardforks::new(Vec::new()),
+                None => {
+                    debug_assert!(
```
procdump, line 219: this will only fire in debug builds which won't happen in production

kstoykov: Yes, this is intentional. It is literally just in case something goes wrong while developing. Otherwise there is a check when a node starts -> resolve_expected_chain_id function

procdump (approved):

PR #172, opened 2026-09-15
feat/schedule record

### LINKED ISSUE #171: Boot gate — refuse a hardfork schedule inconsistent with the executed chain
## Background

At boot the node verifies the datadir's genesis chain-id against the schedule
source selected for the run (built-in `--network` profile or `--config-file`
subnet) and refuses to start on a mismatch. It does **not** verify the
hardfork schedule itself.

An operator who edits fork activation blocks in the schedule source can:

- move an already-activated fork to a different block,
- remove a fork the chain already ran under,
- back-date a new fork into blocks the chain already executed.

Today the node starts anyway, silently re-interpreting the chain's executed
history under a different schedule — different state transitions, rewards,
and precompiles for blocks that already exist.

## Goal

Make the hardfork schedule a pinned, boot-time-verified property of the
datadir, with trust-on-first-use semantics.

kstoykov:
## Summary

- Adds a boot-time gate to `rayls-network node`: the hardfork schedule selected for the
  boot (built-in `--network` profile or `--config-file` subnet) is verified against a new
  `<datadir>/schedule-record.yaml` that pins the schedule the chain's executed blocks ran
  under. Moving, removing, or back-dating a fork whose boundary is at or below the current
  head is refused pre-launch; a boundary strictly in the future is allowed, warned about,
  and re-recorded (that is how agreed schedule updates ship).
- Reads the chain's executed head the same way reth tracks it (the `Finish` stage
  checkpoint in the execution DB) — `CanonicalHeaders` is empty in v2 storage once blocks
  are persisted.
- Datadirs without a record (fresh, or predating this PR) are trusted on first boot and
  start recording from then on; no migration required.

Closes #171

## Surface areas touched

- [x] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

No network upgrade, restart requirement, config migration, or client coordination. Two
behavioral notes for operators:

- Every datadir gains a `schedule-record.yaml` on the next boot. Datadirs that already have
  executed blocks and no record are trusted as-is (boot logs a warning) and the record is
  established from that boot forward.
- From the first recorded boot onward, a boot whose schedule moves a fork boundary that the
  chain has already executed is refused (non-zero exit, pre-launch, no RPC bound). Legitimate
  schedule updates must move boundaries strictly above the current head.

## Test plan

- [x] `cargo test -p rayls-execution-evm` — 135 passed, incl. 14 new unit tests for
      `verify_schedule`/`ScheduleRecord` (identical schedule, executed move refused,
      boundary-at-head refused, boundary-head+1 allowed, back-date refused, removed fork
      refused, chain-id mismatch, unknown fork, `never` ≡ absent, yaml roundtrip,
      `ForkActivation::block`, `ScheduleRecord::activation`).
- [x] `cargo test -p rayls-network-cli` — 26 passed, incl. 7 new integration tests (real
      `init_db` on temp dirs, head faked via a `Finish` checkpoint row): fresh datadir
      records, re-record at the head, executed move refused, future move allowed +
      re-recorded, deleted record re-established, chain-id mismatch, unparseable record.
- [x] `cargo test -p e2e-tests --features dev-single-node-setup --test it schedule_record -- --ignored`
      — full-process e2e on a real dev chain (5 boots): record written on dev boot; tampered
      `--config-file` moving an executed fork refused pre-launch with the fork named; future
      activation allowed and re-recorded; deleted record re-established with
      `as_of_block > 0`.
- [x] `cargo test -p e2e-tests` — default e2e regression: 4 passed / 7 ignored.
- [x] `cargo +nightly-2026-06-24 fmt --all -- --check` and
      `cargo clippy --workspace --all-targets` — no new warnings (the few warnings in
      touched files, e.g. `too_many_arguments` on `RethEnv::new`, pre-date this PR).


procdump (commented):
Two observations:

- As discussed with @bdimitrov-netzine, it is probably more correct to reject --config-file for testnet and mainnet and run them on the built-in schedule only. Two schedule sources for the same network will contradict eventually, and the gate can only judge executed history, not which source is right. Devnet, local and custom networks keep following the file.

- After merge, any developer upgrading the binary on an existing local datadir will be refused at boot: the new fork is active from block 0 in the built-in local table but absent from the record. The refusal is correct, but the message ("boundary changed from never to 0") should tell them the remedy: add the fork to schedule-record.yaml or delete the file so it is re-recorded. Storing never forks in the record explicitly would also let the error distinguish "datadir predates fork Foo" from "Foo was moved from never to 0", e.g.:

```
hardfork 'Foo' is not in the schedule record /data/schedule-record.yaml (written by an
older binary), but the selected schedule activates it at block 0 and the chain is already
at block 5000. Add `Foo: 0` to the record or delete it to re-record on the next boot.
```

procdump:
A fork missing from a subnet's hardforks map is treated as never with no log line. As forks get added to the binary, config files will go stale silently and a node can end up with a fork at never that the rest of the network activates. Since a warning is easy to miss, I'd rather refuse to boot when the file omits a known fork, with an error listing the missing ones, e.g. subnet 'x' in config.yaml does not define hardforks: Foo, Bar. Add each one with a block number or "never". Every binary upgrade that adds a fork then requires a deliberate decision per network, which is what we want.

kstoykov:
> Two observations:
> 
> * As discussed with @bdimitrov-netzine, it is probably more correct to reject --config-file for testnet and mainnet and run them on the built-in schedule only. Two schedule sources for the same network will contradict eventually, and the gate can only judge executed history, not which source is right. Devnet, local and custom networks keep following the file.
> * After merge, any developer upgrading the binary on an existing local datadir will be refused at boot: the new fork is active from block 0 in the built-in local table but absent from the record. The refusal is correct, but the message ("boundary changed from never to 0") should tell them the remedy: add the fork to schedule-record.yaml or delete the file so it is re-recorded. Storing never forks in the record explicitly would also let the error distinguish "datadir predates fork Foo" from "Foo was moved from never to 0", e.g.:
> 
> ```
> hardfork 'Foo' is not in the schedule record /data/schedule-record.yaml (written by an
> older binary), but the selected schedule activates it at block 0 and the chain is already
> at block 5000. Add `Foo: 0` to the record or delete it to re-record on the next boot.
> ```

I suppose that all client network will also have mainnet and testnet. Currently you cannot start a node with flag --network mainnet/testnet and --config-file. Does this cover point 1?

Point 2 - Yes, absolute agree.

procdump:
> > Two observations:
> > 
> > * As discussed with @bdimitrov-netzine, it is probably more correct to reject --config-file for testnet and mainnet and run them on the built-in schedule only. Two schedule sources for the same network will contradict eventually, and the gate can only judge executed history, not which source is right. Devnet, local and custom networks keep following the file.
> > * After merge, any developer upgrading the binary on an existing local datadir will be refused at boot: the new fork is active from block 0 in the built-in local table but absent from the record. The refusal is correct, but the message ("boundary changed from never to 0") should tell them the remedy: add the fork to schedule-record.yaml or delete the file so it is re-recorded. Storing never forks in the record explicitly would also let the error distinguish "datadir predates fork Foo" from "Foo was moved from never to 0", e.g.:
> > 
> > ```
> > hardfork 'Foo' is not in the schedule record /data/schedule-record.yaml (written by an
> > older binary), but the selected schedule activates it at block 0 and the chain is already
> > at block 5000. Add `Foo: 0` to the record or delete it to re-record on the next boot.
> > ```
> 
> I suppose that all client network will also have mainnet and testnet. Currently you cannot start a node with flag --network mainnet/testnet and --config-file. Does this cover point 1?
> 

Can you check - the current conflict prevents --network mainnet --config-file x. It does not prevent --config-file x on a datadir that is already mainnet (through parameters.yaml only). Maybe we should also have this refused?

procdump:
Doesn't the `clap` conflict only block `--network mainnet --config-file` typed together? But.. `--config-file c.yaml --subnet mainnet` on a `mainnet datadir` would still run the file's schedule effectively side-stepping the former protection?

procdump:
Shouldn't this PR take care of what https://github.com/raylsnetwork/axyl/pull/185 does because it's more or less config related as well?

kstoykov:
> Doesn't the `clap` conflict only block `--network mainnet --config-file` typed together? But.. `--config-file c.yaml --subnet mainnet` on a `mainnet datadir` would still run the file's schedule effectively side-stepping the former protection?

`--config-file c.yaml --subnet mainnet` must be a valid one because every custom chain will have 'mainnet' and 'testnet'. The protection that we have in this case is the chain-id check.

Or you idea is to prevent running our chain-ids with --config flag?

kstoykov:
> Shouldn't this PR take care of what #185 does because it's more or less config related as well?

I would prefer a separate PR because this one is large enough already.

procdump:
> Or you idea is to prevent running our chain-ids with --config flag?

Exactly - to protect mainnet/testnet from the public chain to side-step the protection

kstoykov:
> > Or you idea is to prevent running our chain-ids with --config flag?
> 
> Exactly - to protect mainnet/testnet from the public chain to side-step the protection

It has been implemented

PR #173, opened 2026-09-15
ci: publish the node image to public ECR

jairalves:
## Summary

- Publishes the node image to **public** ECR instead of private, so operators can pull it without an AWS account. Part of [SRE-1489](https://parfin.atlassian.net/browse/SRE-1489).
- The promote ladder (devnet → testnet → mainnet) keeps working unchanged in behaviour, but re-tags by a different mechanism — ECR Public has no `batch-get-image`.
- Image is now `public.ecr.aws/rayls/sovereign-axyl:<stage>-vX.Y.Z`.

## Surface areas touched

- [x] CI / build (`.github/workflows/`, `Makefile`)

## Breaking / compatibility

No node, protocol or contract change — this only moves where the image is published.

Two operational consequences:

1. **`AWS_ROLE_ARN` must be repointed** to `arn:aws:iam::785318526575:role/rayls-sovereign-ecr-push` before this merges. Nothing else in this repository uses that secret.
2. **The image moves registry and name.** Anything pulling `<account>.dkr.ecr.us-east-2.amazonaws.com/rayls-stack-node-client` needs to move to `public.ecr.aws/rayls/sovereign-axyl`. Existing tags in the private registry are untouched.

## What changed

Across `build-docker.yml`, `build-docker-replay.yml` and `promote-docker.yml`:

| | from | to |
|---|---|---|
| registry | `<account>.dkr.ecr.us-east-2.amazonaws.com` | `public.ecr.aws/rayls` |
| repository | `rayls-stack-node-client` | `sovereign-axyl` |
| region | `vars.AWS_REGION_PRIVATE` | `us-east-1` — ECR Public has no other endpoint |
| login | default | `registry-type: public` |
| API | `aws ecr …` | `aws ecr-public …` |

The registry has two aliases: `v9x7l0q0` (auto-generated) and `rayls` (custom, primary). We publish under `rayls`. The `sovereign-` prefix keeps these clear of the `rayls-*` repositories the privacy product owns — `rayls-relayer` and `rayls-contracts` are live there on the v2.x line.

### The one non-mechanical change

`promote-docker.yml` re-tagged by reading a manifest with `aws ecr batch-get-image` and re-putting it under the target tag. **ECR Public has no `batch-get-image`**, so that does not port.

It now uses `docker buildx imagetools create`, which does the same job one level up: a registry-side manifest copy that pulls no layers and keeps a multi-arch manifest list intact. Identical content yields an identical digest, and the existing "verify target digest matches source" step still asserts exactly that — so promotion remains a copy of the same bytes, never a rebuild.

## Test plan

- [x] `actionlint` clean on all three workflows (verified it flags a seeded error, so the pass is real)
- [x] IAM simulated against the applied role: `CreateRepository`, `PutImage`, `DescribeImages` allowed on `sovereign-*`; denied on the privacy product's repositories
- [ ] **Not yet run end to end.** No tag has been pushed against this branch, so the first real release is the first real test of the `imagetools` re-tag path.

## Note on #131

[#131](https://github.com/raylsnetwork/axyl/pull/131) (`feat/multi-arch-docker`) touches the same two workflows. Whichever merges second will need a rebase. Multi-arch is deliberately **not** in this PR — for Rust with C dependencies it needs native `ubuntu-24.04-arm` runners and a manifest merge, which is a bigger change than a registry switch.


actionlint
bdimitrov-netzine, line 1: Can we remove the binaries ?

bdimitrov-netzine (dismissed):

bdimitrov-netzine (approved):

PR #179, opened 2026-09-17
feat(fees): wire hybrid performance weights into RewardDistributor payout (opt-in)

### LINKED ISSUE #178: RewardDistributor ignores hybrid performance weights — payout is 100% stake-based
## Summary

`RewardDistributor.distributeRewards()` never reads `ConsensusRegistry.getEpochPerformanceWeights()` — every validator's payout is derived purely from stake and the configured target APY. This was surfaced while investigating a question about whether the ~163 RLS/validator/epoch figure on devnet already accounted for participation/leadership (it doesn't, on either side of the calculation).

Live on devnet, `getEpochPerformanceWeights()` already returns real, differentiated hybrid weights (80% participation / 10% anchor / 10% stake-tier, from #80/#88) — one validator currently sits at roughly a third of its peers' weight — but every validator still receives an identical payout.

## History

This isn't an unbuilt feature — it's a previously-built, deliberately-removed one. The original implementation had `RewardDistributor` distribute `totalRewards * weight_i / totalWeight` directly (with a stake fallback), and it was explicitly removed in #85 ("Remove the leader weights from the rewards distributor calculations") specifically because leader-count-based weighting punished latency-distant validators. `RewardDistributionFairnessTest.t.sol` was written to guard exactly that removal.

The new hybrid weight (80% participation, not just leader-count) is a fairer signal than what was removed, but nobody has re-specified how `RewardDistributor` should consume it. This issue tracks doing that deliberately, as an opt-in mechanism, rather than reintroducing the old behavior unconditionally.

## Related

- #79 (hybrid rollout tracking — ConsensusRegistry side only)
- #80, #88 (hybrid weight computation + activation)
- #85 (removed the original weight-based distribution)

luishsr:
## Summary

`RewardDistributor.distributeRewards()` never read `ConsensusRegistry.getEpochPerformanceWeights()` — payout was 100% stake-derived, even though the hybrid reward model (80% participation + 10% anchor + 10% stake-tier, #80/#88) is live and computing real, differentiated weights on devnet right now. Closes #178.

Adds `performanceWeightBps` (0–10,000) to `RewardDistributor`, **default 0 = byte-identical to current behavior**. When > 0, each validator's stake-derived target is blended with a performance-proportional share of the *same* `totalTarget` — the accumulator pull amount never changes, only the split between validators does. Scoped to the target-APY path only (`_distributeByTarget`); the pre-tier stake fallback (`_distributeByStake`) is untouched.

## ⚠️ Please read before reviewing — this reverses part of #85

This isn't an unbuilt feature — it's a previously-built, **deliberately removed** one. The original implementation had `RewardDistributor` distribute `totalRewards * weight_i / totalWeight` directly. It was explicitly removed in #85 ("Remove the leader weights from the rewards distributor calculations") — `RewardDistributionFairnessTest.t.sol` exists specifically to prove that removal was correct, reproducing "a minority of validators winning far fewer leader rounds than their peers, purely a latency/luck artifact, not a stake or misbehavior signal."

I couldn't find any comment or issue explaining the *deeper* reasoning behind that removal beyond the commit title, and #79/#80/#88 are scoped entirely to the ConsensusRegistry side — nobody has specified how (or whether) RewardDistributor should consume the new hybrid weight. My read: the old weight was pure leader-count (which *is* latency-biased), while the new hybrid weight is 80% participation (certificate inclusion, not leadership — see #80's design notes on why that distinction mattered). That seems like a meaningfully different signal, but since I don't have the original context for #85's removal, I've shipped this as **opt-in and off by default** rather than reintroducing it unconditionally, specifically so this PR can't silently reopen whatever #85 was protecting against. If there's context I'm missing, that's exactly what I'd like this review to surface.

## Devnet verification (real, not simulated)

Deployed to devnet via the existing `UpgradeOps.s.sol` pattern, confirmed `performanceWeightBps()` defaulted to `0` post-upgrade with all existing config intact, then set it to `10_000` and let a real epoch close.

**Before** (all 7 validators, `performanceWeightBps = 0`): every validator held an identical `815353881278538812785` wei pending — confirming pure-stake behavior was unaffected by the upgrade itself.

**After** one epoch with `performanceWeightBps = 10_000`, per-validator reward *this epoch*:

| Validator | Reward this epoch |
|---|---|
| 0x0D77…3e2f | 184.19 RLS |
| 0x3e7f…947D | 166.89 RLS |
| 0xA334…5Ec6 | 166.71 RLS |
| **0xC392…E6c0** | **76.57 RLS** |
| 0xE1a3…0C5 | 180.05 RLS |
| 0xeB72…2441 | 183.71 RLS |
| 0xF841…FffA | 183.37 RLS |

`0xC392…E6c0` (the validator with the lowest recorded hybrid performance weight) earned ~2.3x less than its peers this epoch — matching its weight share. Total distributed ≈ 1,141.49 RLS, matching the pre-change stake-derived `totalTarget` almost exactly, confirming the pot size didn't change, only the split.

## Testing

- `forge build` clean.
- Full suite: **506/506 passing** (502 pre-existing + 4 new). No existing test needed modification — `performanceWeightBps` defaults to 0, so every existing test (including `RewardDistributionFairnessTest.t.sol`, which never calls the new setter) keeps asserting today's behavior unchanged.
- New: `test_distributeRewards_performanceWeightBps_defaultZero_ignoresRecordedWeights` (regression guard), `test_distributeRewards_performanceWeightBps_full_splitsByWeightNotStake` (proves the blend), `test_rewardDistributor_setPerformanceWeightBps_onlyAdmin`, `test_rewardDistributor_setPerformanceWeightBps_boundsCheck`.

## Also fixed while here

A few doc comments (`IConsensusRegistry.sol`, `docs/reward-distribution/README.md`) described the pre-#85 dead-code behavior as if it were live — corrected to describe the actual opt-in mechanism.


luishsr:
Addressed the automated review feedback in 8e60e40:

- **Medium #1** (silent fallback, no on-chain signal): added `PerformanceWeightFetchFailed` event, emitted in the `catch` block — mirrors `AccumulatorTopUpFailed`'s existing pattern.
- **Medium #2** (README diagram implies always-on): fixed to reflect the opt-in default.
- **Medium #3** (stale keeper/ordering comment): fixed — `distributeRewards()` is `onlySystemCall`, runs after `applyIncentives` in the fixed syscall sequence.
- **Low #4**: added a dedicated `InvalidPerformanceWeightBps` error instead of reusing `InvalidApyBps`.
- **Low #5**: documented the rounding-dust-to-Track-B behavior as intentional.
- **Low #6**: added `test_distributeRewards_performanceWeightFetchReverts_fallsBackToStakeAndEmits` covering the previously-untested fallback path.

507/507 tests passing (506 + 1 new). Ready for human review.

PR #181, opened 2026-09-17
fix(rpc): Serve the consensus tip and epoch from a memory watch, and make headers encodable

### LINKED ISSUE #180: [Bug] RPC `rayls_latestHeader` and `rayls_nodeStatus` return stale or failing responses
### What happened?

Two faults:

1. A validator reports header 0 and epoch 0 from its first epoch change onwards, however far the
   chain has advanced. Observers report correctly.
2. A real header cannot be encoded to JSON, so `rayls_latestHeader` returns `-32603 Internal error`.

The "header 0" in fault 1 is the empty header, the only one that encodes, so fault 2 hides fault 1.

### How can it be reproduced?

In `etc/test-network`, `cp .env.example .env`, then set `RAYLS_NETWORK=local` (the file ships
`devnet`, chain id 503, but the script creates chain id 487) and `EPOCH_DURATION=120`. Start the network:

```sh
./local-testnet.sh --start --dev-funds 0x1111111111111111111111111111111111111111
# wait ~2 min for the first epoch change; validator-1 is 8545, the observer 8541
curl -s -X POST -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"rayls_nodeStatus","params":[]}' http://127.0.0.1:8545
curl -s -X POST -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"rayls_latestHeader","params":[]}' http://127.0.0.1:8545
```

### Chain

local (test-network / docker-network)

### Node type

validator (CvvActive)

### Version

1.4.1

### Host OS

Linux (Ubuntu / Debian)

### Storage configuration

```shell

```

### Relevant log output

Snapshot with every node's database at header 2445, epoch 6 (validators 2 and 3 match validator-1):

| node           | `nodeStatus.epoch` | `latestHeader.number` |
|-----------------|------------------------------|----------------------------------|
| validator-1 |                                 0 |                                       0 |
| observer-5 |                                 6 |               `Internal error` |

The validator returns the empty header 0 (fault 1): stale, but it encodes. The observer holds a
real header, so it hits fault 2 — `Internal error`, with `Error serializing response: Error("key
must be a string")` in its log.

## Cause

**Fault 1.** Both values come from `ConsensusBus::last_consensus_header`, updated only when a header
arrives from another node and reset to empty at every epoch change (`epoch_manager/transition.rs`).
A validator builds its own headers, so nothing updates it.

**Fault 2.** Reputation scores are a map keyed by authority id; JSON keys must be text, but an id is
written as a list of 32 numbers. The map is never empty for a real header —
`ReputationScores::new` adds one entry per committee member (`types/primary/reputation.rs:25`) and
every committed sub-dag goes through it (`primary/consensus/bullshark.rs:272`).

ricardopinto:
Fixes two faults in `rayls_latestHeader` and `rayls_nodeStatus`, together because fixing either
alone leaves the pair broken:

1. A validator reports header 0 and epoch 0 from its first epoch change on, however far the chain
   has advanced. Its `last_consensus_header` value is only updated by headers from other nodes and
   is reset every epoch change; a validator builds its own headers, so nothing updates it.
2. A real header fails to encode to JSON (`Internal error`): reputation scores are keyed by
   authority id, and an id was a list of numbers, which cannot be a JSON key.

The empty "header 0" is the only one that encodes, so fault 2 hid fault 1.

## Changes

- Read tip and epoch from a memory watch. Epoch is one after the last *signed* epoch record, or the
  tip's epoch if higher; the unsigned epoch-0 placeholder does not count.
- Observer `is_caught_up` compares the stored tip.
- Authority ids are base58 text in JSON. Binary (BCS, bincode) is byte-for-byte unchanged, so
  stored rows, wire bytes and digests are identical; the old list form still reads.
- Dropped two duplicate tip reads.

## Before / after

Two runs at different points; each table is labelled with the chain's true state.

Without fix, chain at epoch 6, tip 2445:

| node | `nodeStatus.epoch` | `latestHeader.number` |
|---|---|---|
| validator | 0 | 0 |
| observer | 6 | `Internal error` |

With fix, chain at epoch 6, tip 2445:

| node | `nodeStatus.epoch` | `latestHeader.number` |
|---|---|---|
| validator | 6 | 2445 |
| observer | 6 | 2445 |


procdump:
@ricardopinto , I pushed a commit actually reworking the DB idea in favor of a new in memory watch serving the needed data. Can you have a look and test if it works as expected? I would prefer to actually refrain from DB ops here.

ricardopinto:
@procdump I verified and it works great. Your changes also fixed another issue I was tracking regarding epoch transition where, for a brief time after epoch transition, the rpc would return a header that did not exist, like this:
```
rayls_latestHeader -> number 5, leader epoch 0
actual chain tip   -> number 4, leader epoch 1
```

Thank you for your input on this.

procdump:
> @procdump I verified and it works great. Your changes also fixed another issue I was tracking regarding epoch transition where, for a brief time after epoch transition, the rpc would return a header that did not exist, like this:
> 
> ```
> rayls_latestHeader -> number 5, leader epoch 0
> actual chain tip   -> number 4, leader epoch 1
> ```
> 
> Thank you for your inpout on this.

Great then - can you fix the PR title and body so that we can move to merging this one?

ricardopinto:
Changed the title to `fix(rpc): Serve the consensus tip and epoch from a memory watch, and make headers encodable` .

procdump (approved):

kstoykov (approved):

PR #185, opened 2026-09-18
replay: add --config-file/--subnet and bound the hybrid tally walk

bdimitrov-netzine:
--config-file/--subnet installs a network profile before the envs are built, so a replay can apply the schedule a network actually ran instead of the baked-in --chain profile.

The hybrid tally now walks each epoch's ConsensusBlocks rows once with a forward cursor instead of scanning the table tail in reverse per epoch close.

## Summary

-

Closes #

## Surface areas touched

- [ ] Execution / EVM
- [ ] Tests only

## Breaking / compatibility

None.

## Test plan

- [ ]
- [ ]


bin/rayls-replay/src/main.rs
```diff
@@ -5,15 +5,18 @@
...
+    reth_env::{RethCommand, RethConfig, RethEnv},
+    set_active_profile, verify_schedule, NetworkProfile, ScheduleRecord,
+};
+use rayls_infrastructure_config::{Parameters, RaylsDirs};
 use rayls_infrastructure_storage::open_db;
 use rayls_infrastructure_types::{
     rewards::RewardsCounter, Address, Genesis, RaylsNetwork, TaskManager,
 };
-use rayls_middleware_rewards::ConsensusRewardsCounter;
+use rayls_network_cli::schedule::{verify_datadir_chain_id, FileSchedule, SelectedSchedule};
```
procdump: nit: shouldn't we move these functions in rayls_execution_evm::network_profile and skip including rayls_network_cli?

bin/rayls-replay/src/main.rs
```diff
@@ -524,3 +570,63 @@ fn init_tracing(
...
+/// A snapshot taken from a node running the schedule-record boot gate carries
+/// `schedule-record.yaml`, pinning the schedule its executed blocks ran under.
+/// A selection that moves an already-executed fork (or back-dates one into the
+/// executed history) would re-interpret the snapshot's blocks and silently
+/// diverge state, so this refuses it; future boundary moves are reported.
+/// Read-only: replay never writes into the snapshot datadir — the node's gate
+/// re-records, and there is no record here to update. A snapshot without a
+/// record (taken before the feature) is trusted, like the node's no-record
+/// path, minus the write.
+fn verify_snapshot_schedule_record(
```
procdump: nit: is it worth to refactor verify_schedule_record and skip duplicating with this new function?

bin/rayls-replay/src/main.rs
```diff
@@ -227,13 +271,15 @@ async fn run(cli: Cli) -> eyre::Result<()> {
...
+    // hybrid-reward close blocks recompute participation rounds over the snapshot's
+    // consensus DB with a forward, cursor-bounded walk that credits exactly like the
+    // live node's walker but reads each epoch's rows once (see `rewards.rs`; the live
+    // reverse walk scans the whole tail of the table per epoch against a snapshot).
     // ORDERING: this attach must precede `run_replay` below, whose first
     // `install_committee_from_contract` forwards the committee to the walker;
     // `set_committee` only reaches a walker that is already attached.
     if !hybrid_source
-        .attach(RewardsCounter::from_impl(ConsensusRewardsCounter::new(consensus_store.clone())))
+        .attach(RewardsCounter::from_impl(BoundedHybridWalker::new(consensus_store.clone())))
```
procdump: since we don't use ConsensusRewardsCounter in favor of BoundedHybridWalker - let's add some tests?

